PR #30, opened 2021-03-26
Move singletons out of Faster instance

sebastianburckhardt:
Singleton objects clog Faster's object log. Get better perf by storing instance objects separately, and snapshotting them synchronously when creating checkpoing. We store them inside checkpoint commit file.


sebastianburckhardt:
I am observing issues with uploading large files:

CloudBlockBlob.UploadFromStreamAsync: Microsoft.Azure.Storage.StorageException: The specified blob or block content is invalid.

I think it is caused by a partial upload that fails, and then the next upload uses different blocksizes which just does not work. I may need to use unique names for these blobs.

sebastianburckhardt:
I changed the design so that singletons are stored in a separate file inside the cpr-checkpoint folder. This means I do not have to worry about leases, and partial uploads cannot lead to permanently stuck states.

PR #41, opened 2021-05-08
Consolidate organization of performance tests

sebastianburckhardt:
Move tests that were inside separate projects EventProducer/EventConsumer into PerformanceTests, and organized the latter using folders.


sebastianburckhardt:
Continued in #50,

Issue #55, opened 2021-06-21
Netherite breaks due to dll load error


sebastianburckhardt:
## Observed Exception

When recovering a partition after an ungraceful shutdown Netherite can hit this exception:

System.IO.FileNotFoundException: Could not load file or assembly 'System.Threading.Channels, Version=5.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified.
File name: 'System.Threading.Channels, Version=5.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'
   at DurableTask.Netherite.Faster.LogWorker.ReplayCommitLog(Int64 from, StoreWorker worker)
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
   at DurableTask.Netherite.Faster.LogWorker.ReplayCommitLog(Int64 from, StoreWorker worker)
   at DurableTask.Netherite.Faster.StoreWorker.ReplayCommitLog(LogWorker logWorker) in C:\source\durabletask-netherite\src\DurableTask.Netherite\StorageProviders\Faster\StoreWorker.cs:line 418
   at DurableTask.Netherite.Faster.FasterStorage.CreateOrRestoreAsync(Partition partition, IPartitionErrorHandler errorHandler, Int64 firstInputQueuePosition) in C:\source\durabletask-netherite\src\DurableTask.Netherite\StorageProviders\Faster\FasterStorage.cs:line 159

This will effectively prevent Netherite from making any progress and it goes into an infinite recycle.

## Repro
.Net Windows EP3. The DF Template from VS w/ zero changes other than setting extension settings to enable netherite/sql and then the app settings to wire them up.

## Proposed solution

Two things to consider:
1. We nee to solve the dll problem. There is a chance that moving to the latest dependencies could help here.
2. We need to make the problem more observable. I am thinking of (a) forcing early load of this dll, not just at recovery time, and (b) making the load error more observable in the portal, for example as a failure to start function app that creates a visible message.

sebastianburckhardt:
This seems to be a problem with the tooling in Microsoft.NET.Sdk.Functions.

As a temporary workaround, installing an older version (such as 3.0.1) of Microsoft.NET.Sdk.Functions fixes the problem.

shibayan:
[...]

Issue #45, opened 2021-05-21
starting a new orchestration via IDurableClient times out


JackSkylark:
I have a very simple `HttpTrigger` that is attempting to start an orchestration. However, the orchestration never starts and the http request eventually times out. This works as expected in the default durable backend.

```
public class HttpTrigger
{
    private readonly ILogger<HttpTrigger> _logger;

    public HttpTrigger(
        ILogger<HttpTrigger> logger)
    {
        _logger = logger;
    }

    [FunctionName("HttpTrigger")]
    public async Task<IActionResult> RunAsync(
        [HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = "test")] HttpRequest req,
        [DurableClient] IDurableClient durableClient)
    {
        await durableClient.StartNewAsync(
            nameof(ExecutionTestOrchestration));

        return new OkObjectResult("Enqueued");
    }
}
```

sebastianburckhardt:
Are you running this locally? What OS are you using? Do you see any error messages in the console?

JackSkylark:
Yes, I am running locally. Windows 10 (10.0.19042 Build 19042). I never see any error messages.

sebastianburckhardt:
Trying to repro on my own machine, but no success so far. Your code works as it should for me (see details about what I tested below). Not sure where things are going wrong.

Did you try to run the [hello sample](https://microsoft.github.io/durabletask-netherite/#/hello-sample)? Did that work?

To make progress, I may need more information. If you could add the following flags to host.json and send me the console output of the terminal that is running func.exe, it may help:
```
  "logging": {
    "logLevel": {
      // --- the levels below are used to control the Netherite tracing.
      "DurableTask.Netherite": "Debug",
      "DurableTask.Netherite.FasterStorage": "Debug",
      "DurableTask.Netherite.EventHubsTransport": "Debug",
      "DurableTask.Netherite.Events": "Debug",
      "DurableTask.Netherite.WorkItems": "Debug"
    },
  },
```

Another option is to deploy to a premium plan and tell me the region and name of the function app, then I can look at the generated telemetry directly.

## What I tested

 I copied your code into the Hello sample at `durabletask-netherite\samples\Hello\Hello.csproj` and added an orchestration, 
```csharp
  [FunctionName(nameof(ExecutionTestOrchestration))]
        public static Task ExecutionTestOrchestration([OrchestrationTrigger] IDurableOrchestrationContext context, ILogger logger)
        {
[...]

PR #62, opened 2021-07-19
Fix dll load problem and throw exception on 32 bit

sebastianburckhardt:
To avoid dll loading issues as described in #55.


sebastianburckhardt:
Have the runtime bail out with an exception if run on 32 bit, to better handle situation described in #45.

PR #150, opened 2022-04-22
Add Release Pipeline

bachuv:
This PR includes the following changes:
- adding a release pipeline yaml that creates an SBOM (Software Bill of Materials)
- updating .csproj files to include the SBOM in the NuGet package
- updating the VM to the 1ES Hosted VM

### DIFF
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -12,9 +12,7 @@ pool:
     - ImageOverride -equals MMS2022TLS
 
 steps:
-# Start by restoring all the dependencies. This needs to be its own task
-# from what I can tell. We specifically only target DurableTask.AzureStorage
-# and its direct dependencies.
+# Start by restoring all the dependencies.
 - task: DotNetCoreCLI@2
   displayName: 'Restore nuget dependencies'
   inputs:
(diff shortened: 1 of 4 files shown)


azure-pipelines-release.yml
```diff
@@ -0,0 +1,137 @@
...
+
+steps:
+# Start by restoring all the dependencies.
+- task: DotNetCoreCLI@2
+  displayName: 'Restore nuget dependencies'
+  inputs:
+    command: restore
+    verbosityRestore: Minimal
+    projects: |
+      src/DurableTask.Netherite.AzureFunctions/DurableTask.Netherite.AzureFunctions.csproj
```
sebastianburckhardt, line 22: The order of these items should be opposite since the first one depends on the second one.

azure-pipelines-release.yml
```diff
@@ -0,0 +1,137 @@
...
+      src/DurableTask.Netherite/DurableTask.Netherite.csproj
+
+# Build the filtered solution in release mode, specifying the continuous integration flag.
+- task: VSBuild@1
```
sebastianburckhardt, line 26: similarly, this target (Netherite Azure Functions) should come after the next one (Netherite)

sebastianburckhardt (changes requested):
I would prefer some very small changes about the order of projects matching their dependencies.

sebastianburckhardt (approved):

davidmrdavid (approved):
LGTM!

PR #155, opened 2022-04-29
Improve lease management

sebastianburckhardt:
Makes several fixes to the lease management to improve behavior in high-churn situations.

1. When shutting down a partition cleanly, release the lease even if there is an error present, by waiting for lease-using accesses to complete.
2. catch exceptions in a cancellation scenario without generating excessive tracing for them


sebastianburckhardt:
closing this PR as the microsoft-cla bot seems to be glitching

PR #171, opened 2022-06-20
Remove obsolete uses of ConnectionStringResolver

sebastianburckhardt:
This type was causing a compiler warning because it is declared obsolete. Since it appears to be unused, this PR simply removes its uses.


sebastianburckhardt:
I could not get the original PR to work - there were mysterious hangs in the CI that I could not figure out the root cause of. So reverted back to simply disabling the warnings.

PR #178, opened 2022-06-23
New EventHubs performance tests

sebastianburckhardt:
revise design of EventHubs performance test.

Uses a new design:
- the namespace-under-test can contain up to 10 hubs with 32 partitions each
- up to 320 pusher entities can be started, stopped, and queried
- up to 320 puller entities (one per partition) can be started, stopped, and queried
- puller entities use PartitionReceivers to reliably pull events and then either
  a. forward them to an entity, or
  b. start a new orchestration


sebastianburckhardt:
Some things that remain to do:

- [ ] revise host.json changes
- [ ] test autoscaling
- [ ] measure numbers and add to documentation
- [ ] document the design of this test, and how to run it

PR #186, opened 2022-08-05
Updating NetheriteProviderStartup to implement IWebJobsStartup

surgupta-msft:
Updating NetheriteProviderStartup to implement IWebJobsStartup instead of FunctionsStartup

### DIFF
--- a/src/DurableTask.Netherite.AzureFunctions/NetheriteProviderStartup.cs
+++ b/src/DurableTask.Netherite.AzureFunctions/NetheriteProviderStartup.cs
@@ -1,18 +1,15 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
-
-using System.Collections.Concurrent;
-using Microsoft.Azure.Functions.Extensions.DependencyInjection;
-using Microsoft.Azure.WebJobs;
-using Microsoft.Azure.WebJobs.Extensions.DurableTask;
-using Microsoft.Azure.WebJobs.Hosting;
-using Microsoft.Extensions.DependencyInjection;
-
 // Reference: https://docs.microsoft.com/en-us/azure/azure-functions/functions-dotnet-dependency-injection
-[assembly: WebJobsStartup(typeof(DurableTask.Netherite.AzureFunctions.NetheriteProviderStartup))]
+[assembly: Microsoft.Azure.WebJobs.Hosting.WebJobsStartup(typeof(DurableTask.Netherite.AzureFunctions.NetheriteProviderStartup))]
 
 namespace DurableTask.Netherite.AzureFunctions
 {
+    using Microsoft.Azure.WebJobs;
+    using Microsoft.Azure.WebJobs.Extensions.DurableTask;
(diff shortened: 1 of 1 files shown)


src/DurableTask.Netherite.AzureFunctions/NetheriteProviderStartup.cs
```diff
@@ -1,26 +1,28 @@
...
     {
-        public override void Configure(IFunctionsHostBuilder builder)
+        public void Configure(IWebJobsBuilder builder)
         {
 #if !NETCOREAPP2_2
             builder.Services.AddSingleton<IDurabilityProviderFactory, NetheriteProviderFactory>();
 #else
             builder.Services.AddSingleton<IDurabilityProviderFactory, NetheriteProviderPseudoFactory>();
 #endif
+            builder.AddDurableTask();
```
davidmrdavid, line 25: I'm not a blocking reviewer here, but I'm surprised about this change. Why is this here now?

surgupta-msft: I definitely have very limited knowledge about this code. I am just trying to follow pattern used here  https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/DurableTaskWebJobsStartup.cs. Please share all the corrections needed here and I will update the PR.
cc [sebastianburckhardt](https://github.com/sebastianburckhardt)

sebastianburckhardt: I think the `builder.AddDurableTask()` is superfluous here... I would expect this to already be taken care of by the existing startup code in the durable task extension.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderStartup.cs
```diff
@@ -1,26 +1,28 @@
-﻿// Copyright (c) Microsoft Corporation.
+// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Collections.Concurrent;
```
sebastianburckhardt, line 4: In this project we follow the convention of putting the using clauses inside the namespace, as it was the case before. This is obviously not really important, but I prefer it to be consistent across all files in the same project.

sebastianburckhardt (changes requested):
See comments.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderStartup.cs
surgupta-msft: Makes sense. Corrected the code here.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderStartup.cs
surgupta-msft: Done.

sebastianburckhardt (approved):
Looks good to me.

PR #195, opened 2022-10-03
Update to latest Azure Storage Tables client

sebastianburckhardt:
This PR replaces the Azure Table Storage client SDK (v11) with the latest version (v12).

### DIFF
--- a/src/DurableTask.Netherite/Scaling/AzureBlobLoadPublisher.cs
+++ b/src/DurableTask.Netherite/Scaling/AzureBlobLoadPublisher.cs
@@ -18,7 +18,13 @@ namespace DurableTask.Netherite.Scaling
     {
         readonly string taskHubName;
         readonly CloudBlobContainer blobContainer;
-        
+
+        readonly static JsonSerializerSettings serializerSettings = new JsonSerializerSettings() 
+        {
+            TypeNameHandling = TypeNameHandling.None,
+            MissingMemberHandling = MissingMemberHandling.Ignore,
+        };
+
         int? numPartitions;
 
         public AzureBlobLoadPublisher(string connectionString, string taskHubName)
@@ -32,15 +38,9 @@ namespace DurableTask.Netherite.Scaling
 
         public TimeSpan PublishInterval => TimeSpan.FromSeconds(10);
 
-        public Task DeleteIfExistsAsync(CancellationToken cancellationToken)
-        {
-            // not needed since this blob is stored together with the taskhub storage
-            return Task.CompletedTask;
-        }
-
         public Task CreateIfNotExistsAsync(CancellationToken cancellationToken)
         {
-            // not needed since this blob is stored together with the taskhub storage
+            // not needed since the blobs are stored in the taskhub's container
             return Task.CompletedTask;
         }
 
@@ -50,12 +50,36 @@ namespace DurableTask.Netherite.Scaling
             {
                 var blobDirectory = this.blobContainer.GetDirectoryReference($"p{partitionId:D2}");
                 var blob = blobDirectory.GetBlockBlobReference("loadinfo.json");
-                var json = JsonConvert.SerializeObject(loadInfo, Formatting.Indented);
+                var json = JsonConvert.SerializeObject(loadInfo, Formatting.Indented, serializerSettings);
                 return blob.UploadTextAsync(json, cancellationToken);
             }
 
             List<Task> tasks = info.Select(kvp => UploadPartitionInfo(kvp.Key, kvp.Value)).ToList();
-            return Task.WhenAll(tasks);           
+            return Task.WhenAll(tasks);
+        }
+
+        public async Task<T> ReadJsonBlobAsync<T>(CloudBlockBlob blob, bool throwIfNotFound, bool throwOnParseError, CancellationToken token) where T : class
+        {
+            try
+            {
+                var jsonText = await blob.DownloadTextAsync(token).ConfigureAwait(false);
+                return JsonConvert.DeserializeObject<T>(jsonText);
+            }
+            catch (StorageException e) when (!throwIfNotFound && e.RequestInformation?.HttpStatusCode == 404)
+            {
+                // container or blob does not exist
+            }
+            catch (JsonException) when (!throwOnParseError)
+            {
+                // cannot parse content of blob
+            }
+            catch(StorageException e) when (e.InnerException is OperationCanceledException operationCanceledException)
+            {
+                // unwrap the cancellation exception
+                throw operationCanceledException;
+            }
+
+            return null;
         }
 
         public async Task<Dictionary<uint, PartitionLoadInfo>> QueryAsync(CancellationToken cancellationToken)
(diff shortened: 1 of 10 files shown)


src/DurableTask.Netherite/Scaling/AzureTableLoadPublisher.cs
```diff
@@ -13,104 +16,84 @@ namespace DurableTask.Netherite.Scaling
...
         public TimeSpan PublishInterval => TimeSpan.FromSeconds(10);
 
         public async Task DeleteIfExistsAsync(CancellationToken cancellationToken)
         {
-            if (! await this.table.ExistsAsync().ConfigureAwait(false))
+            try
             {
-                return;
+                var tableBatch = new List<TableTransactionAction>();
+                await foreach (var e in this.table.QueryAsync<PartitionInfoEntity>($"PartitionKey eq '{this.taskHubName}'"))
```
jviau, line 35: Consider expression, also pass in cancellation:

```suggestion
                await foreach (var e in this.table.QueryAsync<PartitionInfoEntity>(x => x.PartitionKey == this.taskhubName, cancellationToken: cancellationToken)
```

src/DurableTask.Netherite/Scaling/AzureTableLoadPublisher.cs
```diff
@@ -13,104 +16,84 @@ namespace DurableTask.Netherite.Scaling
...
-        public async Task CreateIfNotExistsAsync(CancellationToken cancellationToken)
+        public Task CreateIfNotExistsAsync(CancellationToken cancellationToken)
         {
-            if (await this.table.ExistsAsync().ConfigureAwait(false))
-            {
-                return;
-            }
-
-            await this.table.CreateAsync().ConfigureAwait(false);
+            return this.table.CreateIfNotExistsAsync();
```
jviau, line 49: pass in cancellation token:
```suggestion
            return this.table.CreateIfNotExistsAsync(cancellationToken);
```

src/DurableTask.Netherite/Scaling/AzureTableLoadPublisher.cs
```diff
@@ -13,104 +16,84 @@ namespace DurableTask.Netherite.Scaling
...
         }
 
         public Task PublishAsync(Dictionary<uint, PartitionLoadInfo> info, CancellationToken cancellationToken)
         {
-            TableBatchOperation tableBatch = new TableBatchOperation();
+            var tableBatch = new List<TableTransactionAction>();
```
jviau, line 54: nit: white space
```suggestion
            var tableBatch = new List<TableTransactionAction>();
```

src/DurableTask.Netherite/Scaling/AzureTableLoadPublisher.cs
```diff
@@ -13,104 +16,84 @@ namespace DurableTask.Netherite.Scaling
...
+            return this.table.SubmitTransactionAsync(tableBatch, cancellationToken);
         }
 
         public async Task<Dictionary<uint, PartitionLoadInfo>> QueryAsync(CancellationToken cancellationToken)
         {
-            var query = new TableQuery<PartitionInfoEntity>().Where(TableQuery.GenerateFilterCondition("PartitionKey", QueryComparisons.Equal, this.taskHubName));
-            TableContinuationToken continuationToken = null;
             Dictionary<uint, PartitionLoadInfo> result = new Dictionary<uint, PartitionLoadInfo>();
-            do
+            await foreach (var e in this.table.QueryAsync<PartitionInfoEntity>($"PartitionKey eq '{this.taskHubName}'"))
```
jviau, line 65: Same as above, expression and cancellation token:

```suggestion
            await foreach (var e in this.table.QueryAsync<PartitionInfoEntity>(x => x.PartitionKey == this.taskhubName, cancellationToken: cancellationToken)
```

src/DurableTask.Netherite/Scaling/AzureTableLoadPublisher.cs
```diff
@@ -13,104 +16,84 @@ namespace DurableTask.Netherite.Scaling
...
             {
-                var batch = await this.table.ExecuteQuerySegmentedAsync<PartitionInfoEntity>(query, continuationToken, null, null, cancellationToken).ConfigureAwait(false);
-                foreach (var e in batch)
+                int.TryParse(e.CachePct, out int cachePct);
```
jviau, line 67: is it fine if this and the double fail to parse?

jviau (approved):

src/DurableTask.Netherite/Scaling/AzureTableLoadPublisher.cs
sebastianburckhardt: I did not know this was supported, will definitely use that (as long as we can be sure the query is compiled correctly).

src/DurableTask.Netherite/Scaling/AzureTableLoadPublisher.cs
sebastianburckhardt: These days I try to remove the ConfigureAwait(false) noise. All my code is running in tasks on the thread pool, there is no implicit synchronization context to worry about.

sebastianburckhardt: Possibly so. I will take another look. Usually I don't bother because most of my code is only ever called from thread pool tasks, but this may be called from some place that has an implicit synchronization context.

src/DurableTask.Netherite/Scaling/AzureTableLoadPublisher.cs
sebastianburckhardt: There is no good way to log errors from this code, so the calling code has to catch and report exceptions. Same is true for other exceptions.

sebastianburckhardt: yeah, the exception policy is not very consistent actually. Also, this class should really have unit tests, I think there are bugs here.

src/DurableTask.Netherite/Scaling/AzureBlobLoadPublisher.cs
```diff
@@ -50,44 +50,96 @@ Task UploadPartitionInfo(uint partitionId, PartitionLoadInfo loadInfo)
...
+                // container or blob does not exist
+            }
+            catch (JsonException) when (!throwOnParseError)
+            {
+                // cannot parse content of blob
+            }
+            catch(StorageException e) when (e.InnerException is OperationCanceledException operationCanceledException)
+            {
+                // unwrap the cancellation exception
+                throw operationCanceledException;
```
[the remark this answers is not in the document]
sebastianburckhardt: I don't think I know what that means. But it probably does not matter since cancellations are not "real" exceptions, i.e. nothing went wrong and no one is going to care where it was thrown.

src/DurableTask.Netherite/Scaling/AzureBlobLoadPublisher.cs
```diff
@@ -50,44 +50,96 @@ Task UploadPartitionInfo(uint partitionId, PartitionLoadInfo loadInfo)
...
                 var blobDirectory = this.blobContainer.GetDirectoryReference($"p{partitionId:D2}");
                 var blob = blobDirectory.GetBlockBlobReference("loadinfo.json");
-                var json = JsonConvert.SerializeObject(loadInfo, Formatting.Indented);
+                var json = JsonConvert.SerializeObject(loadInfo, Formatting.Indented, serializerSettings);
                 return blob.UploadTextAsync(json, cancellationToken);
             }
 
             List<Task> tasks = info.Select(kvp => UploadPartitionInfo(kvp.Key, kvp.Value)).ToList();
-            return Task.WhenAll(tasks);           
+            return Task.WhenAll(tasks);
```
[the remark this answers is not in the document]
sebastianburckhardt: yes, I forgot about that. Will revise.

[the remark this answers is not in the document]
sebastianburckhardt: I don't think I need to use `ConfigureAwait` on these lines since I am not using an await but returning the Task. Still, I get your point. There are other instances.

PR #197, opened 2022-10-13
Add support for authentication using an Azure.Core.TokenCredential

sebastianburckhardt:
The following mechanisms are introduced:

1. in NetheriteOrchestrationServiceSettings, we replaced "ResolvedConnectionString" with a new class "ConnectionInfo". This is a class that contains enough information to construct objects needed for connections (e.g. SDK clients), such as connection strings, OR token credentials and endpoint URI information. We use separate ConnectionInfo members for all types of connections (table storage, blob storage, page blob storage, eventhubs).

2. Implement ConnectionInfo extension methods that can construct various SDK objects (v12 table clients, v11 storage accounts, EventHubClient, EventProcessorHost) and authenticate HTTP requests

3. Create a new abstract class 'ConnectionResolver' that resolves the required ConnectionInfo objects before starting the orchestration service. A `ConnectionResolver` must be run on the settings object before the service can be started.  DTFx clients call this explicitly (since they are constructing a settings object in code). DF clients must inject a connection resolver during startup, since they cannot access the settings object directly. This class has several implementations suited for different use cases, and can also be implemented by users directly if desired:

    * `ConnectionNameToConnectionStringResolver` represents the currently implemented way in which names are resolved to connection strings. This is used by default, if the user does not inject a different resolver
    *  `SimpleCredentialResolver` has a constructor that takes an `Azure.Core.TokenCredential` and two names (for storage account and event hubs namespace). It is intended for convenient use by DTFx applications.
    * `TokenCredentialResolver` is an abstract class whose constructor takes an `Azure.Core.TokenCredential` and that must be overriden to map connection names to resource names (storage account and event hub name). It is intended to be used with DF's current mechanism where connection names are a application-visible concept (e.g. you can specify a particular connection to be used for a DurableClient).

5. Added two samples that demonstrate how to use the new mechanism for DF and DTFx, respectively.


src/DurableTask.Netherite/Connections/ConnectionInfo.cs
```diff
@@ -0,0 +1,152 @@
...
+            switch (resourceType)
+            {
+                case ConnectionResolver.ResourceType.EventHubsNamespace:
+                    {
+                        return new ConnectionInfo()
+                        {
+                            ResourceName = name,
+                            ConnectionString = null,
+                            TokenCredential = tokenCredential,
+                            HostName = $"{name}.servicebus.windows.net",
```
jviau, line 112: if we want to support Netherite in other Azure clouds we will need some follow up work for cloud-specific host names. Or can this be adjusted to take in the full hostname instead of name?

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
```diff
@@ -88,58 +103,74 @@ public override string ToString()
...
 #endif
-            return $"NetheriteOrchestrationService on {this.configuredTransport}Transport and {this.configuredStorage}Storage, {configuration} build";
+            return $"NetheriteOrchestrationService on {this.Settings.TransportChoice}Transport and {this.Settings.StorageChoice}Storage, {configuration} build";
         }
 
         /// <summary>
         /// Creates a new instance of the OrchestrationService with default settings
         /// </summary>
-        public NetheriteOrchestrationService(NetheriteOrchestrationServiceSettings settings, ILoggerFactory loggerFactory)
+        public NetheriteOrchestrationService(NetheriteOrchestrationServiceSettings settings, ILoggerFactory loggerFactory, IServiceProvider serviceProvider = null)
```
jviau, line 112: It is preferable to take in the exact dependencies you need (to follow IoC / DI) vs `IServiceProvider`.

As with the other ctor comment, I recommend avoiding a breaking change by adding a separate ctor which takes in the new params.

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationServiceSettings.cs
```diff
@@ -28,22 +29,11 @@ public class NetheriteOrchestrationServiceSettings
...
+        /// Pseudo-connection-strings "Memory" or "SingleHost" can be used to configure
+        /// in-memory emulation, or single-host configuration, respectively.
         /// </summary>
         public string EventHubsConnectionName { get; set; } = "EventHubsConnection";
 
-        /// <summary>
-        /// The resolved storage connection string. Is never serialized or deserialized.
-        /// </summary>
-        [JsonIgnore]
-        public string ResolvedStorageConnectionString { get; set; }
```
jviau, line 39: There are several breaking changes in this file. Are we okay with that? If we follow semver, that would require a major version rev.

The alternative is to try and do additive changes only - retaining old functionality as is and only adding it what you need for the new functionality.

jviau (commented):
It is great to see AAD auth being supported. However, I am a bit concerned with the number of breaking changes here, especially since Netherite is GA. These changes will require a 2.0 release and do we want to do that so soon? Is there a way to accomplish this in a pure additive non-breaking fashion?

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
```diff
@@ -114,15 +119,31 @@ NetheriteOrchestrationServiceSettings GetNetheriteOrchestrationServiceSettings(s
                 netheriteSettings.HubName = taskHubNameOverride;
             }
 
+            // connections for Netherite are resolved either via an injected custom resolver, or otherwise by resolving connection names to connection strings
+            var connectionResolver = this.serviceProvider.GetService<DurableTask.Netherite.ConnectionResolver>()
```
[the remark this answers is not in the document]
sebastianburckhardt: Yes, I will add that.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
```diff
@@ -48,6 +51,7 @@ public class NetheriteProviderFactory : IDurabilityProviderFactory
         public NetheriteProviderFactory(
             IOptions<DurableTaskOptions> extensionOptions,
             ILoggerFactory loggerFactory,
+            IServiceProvider serviceProvider,
```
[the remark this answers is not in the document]
sebastianburckhardt: Yep, constructor overloading sounds like a good way to avoid a breaking change.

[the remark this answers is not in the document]
sebastianburckhardt: Is this really a breaking change? Adding an optional argument should not break any existing uses of the constructor.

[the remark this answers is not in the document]
sebastianburckhardt: I agree though that it does seem like a good idea to not force users into using an IServiceProvider which could be inconvenient if they are not already using DI.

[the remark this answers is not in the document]
jviau: Yes, it is a runtime breaking change - existing apps that drop in the version of this assembly will encounter `MethodMissingException` as this is a different ctor signature at runtime.

src/DurableTask.Netherite.AzureFunctions/IConnectionInfoFactory.cs
```diff
@@ -0,0 +1,31 @@
...
+    using System.Text;
+    using Azure.Identity;
+    using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+
+#if !NETCOREAPP2_2           
+
+    /// <summary>
+    /// Allows customization of what settings to use for what connection. Can be used to specify TokenCredentials. 
+    /// </summary>
+    public abstract class ConnectionSettingsCustomizer
```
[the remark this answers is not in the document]
sebastianburckhardt: Looks like it is dead code left over from some earlier version. Will remove it.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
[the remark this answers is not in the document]
sebastianburckhardt: This constructor is never called by any client code, it is only accessible via DI.
So this is not actually a breaking change.

src/DurableTask.Netherite/DurableTask.Netherite.csproj
```diff
@@ -51,6 +51,7 @@
   
   <ItemGroup>
     <PackageReference Include="Azure.Data.Tables" Version="12.6.1" />
+    <PackageReference Include="Azure.Identity" Version="1.6.0" />
```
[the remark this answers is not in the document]
sebastianburckhardt: Nice.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
```diff
@@ -8,11 +8,13 @@ namespace DurableTask.Netherite.AzureFunctions
     using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Threading;
+    using Azure.Identity;
```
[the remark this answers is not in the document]
sebastianburckhardt: will remove.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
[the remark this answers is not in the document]
sebastianburckhardt: I was careful to avoid a breaking change for AF users.

I was not careful for DF users though, so I should add some overloads to keep things compatible.

src/DurableTask.Netherite.AzureFunctions/TokenCredentialResolver.cs
```diff
@@ -0,0 +1,80 @@
...
+                case ResourceType.EventHubsNamespace:
+                    string eventHubsNamespaceName = this.GetEventHubsNamespaceName(connectionName);
+                    if (string.IsNullOrEmpty(eventHubsNamespaceName))
+                    {
+                        throw new ArgumentException("GetEventHubsNamespaceName returned invalid result");
+                    }
+                    return ConnectionInfo.FromTokenCredential(this.tokenCredential, eventHubsNamespaceName, recourceType);
+
+                default:
+                    throw new NotImplementedException("unknown resource type");
```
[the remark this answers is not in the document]
sebastianburckhardt: ok.

src/DurableTask.Netherite.AzureFunctions/TokenCredentialResolver.cs
```diff
@@ -0,0 +1,80 @@
...
+    using System.Text;
+    using Azure.Identity;
+    using Microsoft.Extensions.Logging.Abstractions;
+
+#if !NETCOREAPP2_2
+
+    /// <summary>
+    /// Resolves connections using a token credential and a mapping from connection names to resource names.
+    /// </summary>
+    public abstract class CredentialBasedConnectionNameResolver : DurableTask.Netherite.ConnectionResolver
```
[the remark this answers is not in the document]
sebastianburckhardt: ok.

src/DurableTask.Netherite/Connections/ConnectionInfo.cs
sebastianburckhardt: Yes, the hardcoded string here is only meant to be a convenient default.
The connection resolver can set this any other hostname.

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
sebastianburckhardt: I don't prefer using a list of dependencies. It means we have to add/modify constructor overloads every time the list changes which is really inconvenient. Also, this does not work here since not all of the types are public.

How is this a breaking change?  the argument is optional.

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationServiceSettings.cs
sebastianburckhardt: I can see how much work is required to maintain compatibility. Perhaps keeping some deprecated ResolvedConnectionString fields is enough.

Note that this affects only DTFx users, for DF users there are no breaking changes.

sebastianburckhardt: I will add a compatibility shim so there are no breaking changes.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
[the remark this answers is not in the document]
jviau: Many breaking changes can be runtime only. This will break if I take an app compiled against the previous version of DurableTask.Netherite and drop in the new dll. When following semver, you should always be able to drop in a minor version rev _without needing to recompile your application._

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
jviau: As mentioned in the other comment, it is breaking because code compiled against the previous dll will fail to find this constructor if the new version is dropped in place.

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationServiceSettings.cs
jviau: Yeah, avoiding breaking changes is very tricky. Here is a list of what is and is not allowed: https://learn.microsoft.com/en-us/dotnet/core/compatibility/

jviau: To avoid breaking changes in this PR, I suggest making sure everything is purely additive. Do not remove or change existing method signatures or properties - only add new ones. The implementations then need to make decisions at runtime for which property to use. You _can_ update the getting/setting implementations though. So, an example of what you are trying to update could be like:

Before:
``` CSharp
public string OldConnectionString { get; set; }
``` 

After:
``` CSharp
public ComplexConnectionType { get; set; }

public string OldConnectionString
{
    get => ComplexConnectionType?.ConnectionString;
    set => ComplexConnectionType = new ComplexConnectiontype(value); // also need some logic around going from non-null to null
}

public class ComplexConnectionType
{
    public ComplexConnectionType(string connectionString) { } // a ctor for the connection string scenario

    public ComplexConnectionType(TokenCredential credential) { } // a ctor for the new recommended way. Just an example implementation - format this however you need.
```

sebastianburckhardt: We can then remove the shim when we go to 2.0

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
sebastianburckhardt: Ok, I will create an overload then.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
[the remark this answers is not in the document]
sebastianburckhardt: Ok, got it. It means I cannot change method signatures.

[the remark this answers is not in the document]
sebastianburckhardt: I think DI is based on runtime reflection, so I think it should work. I can test that.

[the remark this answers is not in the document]
sebastianburckhardt: > When following semver, you should always be able to drop in a minor version rev without needing to recompile your application.

Supporting update without recompilation for _any_ possible situation seems like an unnecessarily high bar. I can see this being important for some projects, but it seems like overkill in this context if applied as an abstract rule. IMO it is enough to consider the supported ways in which our APIs may be used, and make sure to support updates for those supported scenarios. For example, we can assume that Netherite.AzureFunctions.dll is only used from Microsoft.Azure.WebJobs.Extensions.DurableTask.dll.

Also, I cannot imagine that there are many users that need to update Netherite without recompiling. That seems like a very fringe scenario. From what I have seen, any function app deployments are being recompiled right before being deployed.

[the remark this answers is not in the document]
sebastianburckhardt: I have manually tested on the sample function app that I can drop in the new dlls (DurableTask.Netherite.AzureFunctions.dll and DurableTask.Netherite.dll) and the app still runs, without recompilation.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
```diff
@@ -114,15 +118,31 @@ NetheriteOrchestrationServiceSettings GetNetheriteOrchestrationServiceSettings(s
                 netheriteSettings.HubName = taskHubNameOverride;
             }
 
+            // connections for Netherite are resolved either via an injected custom resolver, or otherwise by resolving connection names to connection strings
+            var connectionResolver = this.serviceProvider?.GetService<DurableTask.Netherite.ConnectionResolver>()
```
jviau, line 122: I think there is an opportunity to follow DI/IoC with the connection resolver, but still make it ultimately "optional" for customers. The idea would be to move this kind of logic to the DI setup code. I'll have to take a closer look at what existing DI setup code Netherite has, but I think we can definitely get something working.

For example, is `INameResolver` even used directly here, or is it only used within the `ConnectionResolver`? If it is only used within the connection resolver, we can do the following:

1. Add a 2nd constructor which takes in `ConnectionResolver` and _no_ `INameResolver`.
2. The ctor which still has `INameResolver` just calls that new ctor with the `new ConnectionNameToConnectionStringResolver((name) => nameResolver.Resolve(name))` as the connection resolver.
    - No longer have a `INameResolver` field. Which is why it is important INameResolver is not used directly anymore. If we still need it directly, will need to rethink this aproach.
4. Add `[ActivatorUtilitiesConstructor]` to the new connection resolver ctor (so it is used for DI).
5. Update the DI setup code to use `TryAddSingleton<ConnectionResolver>` with a factory which will pull `INameResolver` from the service container and return `new ConnectionNameToConnectionStringResolver((name) => nameResolver.Resolve(name))` as well.
    - at this point, maybe a constructor for `ConnectionNameToConnectionStringResolver` which takes in `INameResolver` may be nice.

What we now have is no observable difference between old and new. However, customers can now inject `ConnectionResolver` to the DI container and it will be consumed as expected, but we have maintained DI/IoC principals.

src/DurableTask.Netherite/Connections/ConnectionInfo.cs
```diff
@@ -0,0 +1,148 @@
...
+// Licensed under the MIT License.
+
+namespace DurableTask.Netherite
+{
+    using System;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Storage;
+
+    /// <summary>
+    /// Internal abstraction used for capturing connection information and credentials.
```
jviau, line 11: Considering this type is `public`, should "Internal" be removed from the comment?

src/DurableTask.Netherite/Connections/ConnectionInfoExtensions.cs
```diff
@@ -0,0 +1,219 @@
...
+    using Azure.Core;
+    using System.Runtime.CompilerServices;
+    using Microsoft.Azure.EventHubs.Processor;
+    using Newtonsoft.Json.Serialization;
+    using DurableTask.Netherite.Faster;
+
+    /// <summary>
+    /// Utilities for constructing various SDK objects from a connection information.
+    /// </summary>
+    public static class ConnectionInfoExtensions
```
jviau, line 27: Does this need to be `public`? With all the breaking change discussions in this PR, I recommend only making types `public` if they are expected to be consumed by end users.

If they need to be `public` for cross-assembly usage within this repo, we can use [internals-visible-to](https://github.com/microsoft/durabletask-dotnet/blob/main/eng/targets/InternalsVisibleTo.targets).

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
```diff
@@ -131,7 +151,8 @@ NetheriteOrchestrationServiceSettings GetNetheriteOrchestrationServiceSettings(s
                 netheriteSettings.CacheOrchestrationCursors = false; // cannot resume orchestrations in the middle
             }
 
-            netheriteSettings.Validate((name) => this.nameResolver.Resolve(name));
+            // validate the settings and resolve the connections
+            netheriteSettings.Validate(connectionResolver);
```
jviau, line 155: If you are interested, `IOptions` has a validation pattern to it. Probably better left to a different PR (time and priority permitting). But I find it helps with encapsulation, so you can take in `IOptions<T>` and trust that object is fully valid by the time it is added here.

src/DurableTask.Netherite/Connections/ConnectionNameToConnectionStringResolver.cs
```diff
@@ -0,0 +1,76 @@
...
+    using System.Collections.Generic;
+    using System.Data;
+    using System.Text;
+    using Microsoft.Azure.Storage;
+    using Microsoft.Extensions.Logging.Abstractions;
+
+    /// <summary>
+    /// Resolves connection names by using a mapping from connection names to connection strings.
+    /// </summary>
+    public class ConnectionNameToConnectionStringResolver : ConnectionResolver
```
jviau, line 16: In addition to my other suggestion around maintaining IoC pattern with the factory, maybe this type could be made `internal`. I like to keep public API surface to only what is necessary. Once something is public, it is hard to take it back.

src/DurableTask.Netherite/Connections/CredentialShim.cs
```diff
@@ -0,0 +1,52 @@
...
+    using Azure.Core;
+    using Microsoft.Azure.Storage.Auth;
+    using System.Threading.Tasks;
+    using System.Threading;
+
+    /// <summary>
+    /// A utility class for constructing a <see cref="Microsoft.Azure.Storage.Auth.TokenCredential"/> 
+    /// from a <see cref="Azure.Core.TokenCredential"/>.
+    /// </summary>
+    public static class CredentialShim
```
jviau, line 18: Consider making this `internal`.

src/DurableTask.Netherite/Connections/EventHubsUtil.cs
```diff
@@ -24,9 +24,9 @@ public static class EventHubsUtil
         /// <param name="eventHubName">The name of the event hub.</param>
         /// <param name="partitionCount">The number of partitions to create, if the event hub does not already exist.</param>
         /// <returns>true if the event hub was created.</returns>
-        public static async Task<bool> EnsureEventHubExistsAsync(string connectionString, string eventHubName, int partitionCount)
+        public static async Task<bool> EnsureEventHubExistsAsync(ConnectionInfo info, string eventHubName, int partitionCount, CancellationToken cancellationToken)
```
jviau, line 27: This is technically a public API contract, but I get this was never meant for public use. Use your judgement here - do you make it separate methods to avoid breaking change or not?

In the next major version rev I definitely recommend evaluating public API surface and making all utilities like this and other implementation details classes `internal`.

src/DurableTask.Netherite/OrchestrationService/NetheriteConfigurationException.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Text;
+
+    /// <summary>
+    /// An exception that indicates configuration errors for Netherite.
+    /// </summary>
+    [Serializable()]
+    public class NetheriteConfigurationException : System.Exception
```
jviau, line 14: As mentioned, the options pattern does have a validation exception type in it: https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.options.optionsvalidationexception?view=dotnet-plat-ext-6.0

With some refactoring, could leverage the options pattern more directly to avoid having to recreate this ourselves.

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
```diff
@@ -88,58 +103,82 @@ public override string ToString()
...
         /// </summary>
         public NetheriteOrchestrationService(NetheriteOrchestrationServiceSettings settings, ILoggerFactory loggerFactory)
+            : this(settings, loggerFactory, null)
+        {
+        }
+        
+        /// <summary>
+        /// Creates a new instance of the OrchestrationService with default settings
+        /// </summary>
+        public NetheriteOrchestrationService(NetheriteOrchestrationServiceSettings settings, ILoggerFactory loggerFactory, IServiceProvider serviceProvider)
```
jviau, line 120: As with the `NetheriteProviderFactory`, I recommend using DI/IoC over service-locator.

jviau (commented):
Some comments for now. These are all up to you if you want to follow them or not.

src/DurableTask.Netherite/Connections/EventHubsUtil.cs
sebastianburckhardt: Yep, I was not particularly aware of how ANY public API surface becomes a liability. We can definitely retire some of these in the next major version change.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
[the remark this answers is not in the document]
sebastianburckhardt: > I also recommend to split these changes into a different constructor. This will avoid breaking changes and also if there is an either/or option (provide creds this way or a different way) the two ctors can represent the two different ways.

how does DI know which constructor to use if there is more than one?

src/DurableTask.Netherite/Connections/ConnectionInfoExtensions.cs
sebastianburckhardt: These methods are useful if you want to slightly change something (e.g. the host URL) without having to rewrite all of it.

src/DurableTask.Netherite/Connections/ConnectionInfo.cs
sebastianburckhardt: yes

src/DurableTask.Netherite/Connections/SimpleCredentialResolver.cs
```diff
@@ -0,0 +1,74 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Text;
+    using Azure.Core;
+    using Microsoft.Extensions.Logging.Abstractions;
+
+    /// <summary>
+    /// Resolves connections using a token credential, a storage account name, and an eventhubs namespace name.
+    /// </summary>
+    public class SimpleCredentialResolver : ConnectionResolver
```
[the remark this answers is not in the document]
sebastianburckhardt: Yes, that seems like a good way to do it. Will adjust.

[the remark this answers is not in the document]
sebastianburckhardt: Makes sense. Though I prefer to keep the actual classes in their own file. Just made them internal and added public factory methods to the base class.

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
sebastianburckhardt: I understand that the service locator may not always be the preferred solution, but I think it is appropriate in this situation.

- Not all of the services are public, so it is not just a simple case of listing them in the constructor argument. 
- Constructors that list all the dependencies create maintenance complexity. In some sense, this is the problem DI is solving in the first place so it seems sad to give up on that. Also, I really dislike having lots of overloads, it makes the code hard to read.
 
Using a service locator instead helps to keep the constructor stable because new dependencies can be found without having to be listed explicitly. 

From what I understand, the danger of the service locator pattern is that the dependency ordering is not visible to the DI startup logic, so it cannot correctly order dependencies during startup. But that is not an issue here because the orchestration service is never constructed at that point. Only the factories are constructed at startup time.

src/DurableTask.Netherite/OrchestrationService/NetheriteConfigurationException.cs
sebastianburckhardt: I will leave the refactoring of the settings into an options pattern up to a future PR (perhaps for a major version change). It is not a goal of this PR to align Netherite with best DI practices.

src/DurableTask.Netherite/Connections/CredentialShim.cs
sebastianburckhardt: yes.

src/DurableTask.Netherite/Connections/ConnectionNameToConnectionStringResolver.cs
sebastianburckhardt: yes.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
sebastianburckhardt: I will leave that to a future refactoring in a separate PR.

src/DurableTask.Netherite/Connections/ConnectionInfoExtensions.cs
sebastianburckhardt: I think they can be internal, yes.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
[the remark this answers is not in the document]
sebastianburckhardt: > how does DI know which constructor to use if there is more than one?

I see you have answered this below. There is an attribute for it.

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationServiceSettings.cs
sebastianburckhardt: There are no more breaking changes here.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
sebastianburckhardt: > However, customers can now inject ConnectionResolver to the DI container 

I don't quite understand how they can inject a ConnectionResolver if we already injected our own during our own setup

sebastianburckhardt: Empirically, it does seem to work... 
- The TokenCredentialDF.FunctionStartup calls `builder.Services.AddSingleton<DurableTask.Netherite.ConnectionResolver, MyConnectionResolver>();`
- The DurableTask.Netherite.AzureFunctions.NetheriteProviderStartup calls `builder.Services.TryAddSingleton<ConnectionResolver, NameResolverBasedConnectionNameResolver>();`

and magically the right things happen. Did I just get lucky or is the DI smart enough to know which one should take precedence?

sebastianburckhardt: > If we still need it directly, will need to rethink this aproach

I still need it, but that does not really change anything I think. I just leave it there, and go ahead with the rest of the suggestions.

sebastianburckhardt:
There are no more breaking changes, and I have revised some of the DI patterns as recommended. However, I kept the service-locator pattern in one specific situation because I did not like the alternatives. I also added README files to the samples so users understand better what is being demonstrated and how to run them.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
jviau: What is happening is `builder.Services.TryAddSingleton<ConnectionResolver, NameResolverBasedConnectionNameResolver>();` will first check a service for `ConnectionResolver` is already registered, only adding `NameResolverBasedConnectionNameResolver` if no `ConnectionResolver` is already registered. This gives end users control over the which instance is registered: if they run before you, they can just call `AddSingleton`. If they run after, they will first need to remove your service registration then add their own (this is possible, `IServiceCollection` is a fully mutable collection with add/remove semantics).

src/DurableTask.Netherite/Connections/ConnectionInfoExtensions.cs
jviau: As helpful as providing extensions like this to customers is, I advise against it. Public API surface should be tightly controlled for maintainability's sake - each and every public API needs to have a good justification to be public. If customers think this would be helpful, they can copy the code into their own project.

src/DurableTask.Netherite/Connections/SimpleCredentialResolver.cs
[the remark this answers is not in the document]
jviau: That approach is completely fine as well. I keep them in their own file as well - but I make the outer class `partial` and move the nested class to its own file. So, it would be `ConnectionResolver.SimpleCredentialResolver.cs`.

But not need for that if you already made it `internal`.

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
```diff
@@ -46,18 +47,39 @@ public class NetheriteProviderFactory : IDurabilityProviderFactory
         public const string ProviderName = "Netherite";
         public string Name => ProviderName;
 
-        // Called by the Azure Functions runtime dependency injection infrastructure
+        /// <summary>
+        /// This constructor should not be called directly. It is here only to avoid breaking changes.
+        /// </summary>
+        [Obsolete]
```
jviau, line 53: Obsolete'ing APIs has some differing viewpoints. .NETs policy is as follows:

> An API that shipped in a long-term support (LTS) release must be obsoleted in the subsequent LTS release before it can be removed. In rare cases, exceptions are made to obsolete an API before the subsequent LTS release based on business needs. All obsoletions are documented and communicated to customers.

It takes _two_ LTS releases to remove a public API! First one marks it obsolete, second removes. This is because _technically_ adding `[Obsolete]` can be viewed as a compilation breaking change if the project has `WarnAsError=true`.

Buuuut, I think in this case you are fine. This is just more of a fun FYI to express how "_fun_" maintaining public APIs is 😆

src/DurableTask.Netherite/Connections/ConnectionResolver.cs
```diff
@@ -58,5 +58,26 @@ public enum ResourceType
         /// <param name="storageChoice">The storage layer to use.</param>
         /// <param name="transportChoice">The transport layer to use.</param>
         public abstract void ResolveLayerConfiguration(string connectionName, out StorageChoices storageChoice, out TransportChoices transportChoice);
+
+        /// <summary>
+        /// Creates a connection resolver for a given connection string lookup function.
+        /// </summary>
+        /// <param name="connectionStringLookup">A function that maps connection names to connection strings.</param>
+        public static ConnectionResolver FromConnectionNameToConnectionStringResolver(Func<string, string> connectionStringLookup)
```
jviau, line 66: Is there a more concise name we can use? We can at least drop `Resolver` suffix, as that is already expressed in the type ("ConnectionResolver"). For a more concise name, maybe `FromNamedConnectionString`?

src/DurableTask.Netherite/Connections/ConnectionResolver.cs
```diff
@@ -58,5 +58,26 @@ public enum ResourceType
...
+        {
+            return new ConnectionNameToConnectionStringResolver(connectionStringLookup);
+        }
+
+        /// <summary>
+        /// Creates a connection resolver given a token credential, a storage account name, and an event hubs namespace name.
+        /// Applies the default suffixes (*.core.windows.net for storage accounts and *.servicebus.windows.net for event hubs).
+        /// </summary>
+        /// <param name="tokenCredential">The token credential to use.</param>
+        /// <param name="storageAccountName">The name of the storage account, or null if using in-memory emulation.</param>
```
jviau, line 76: Does in-memory require a credential at all? Should an explicit separate method be added instead `ForInMemory()`?

src/DurableTask.Netherite/Connections/ConnectionResolver.cs
```diff
@@ -58,5 +58,26 @@ public enum ResourceType
...
+        /// <param name="eventHubNamespaceName">The name of the event hub namespace, or null if using the singlehost configuration.</param>   
+        public static ConnectionResolver FromTokenCredentialAndResourceNames(Azure.Core.TokenCredential tokenCredential, string storageAccountName = null, string eventHubNamespaceName = null)
```
jviau, line 78: What do you think of dropping "AndResourceNames"? `FromTokenCredential(...)` sounds fine to me, even including the resource name parameters.

jviau (approved):

src/DurableTask.Netherite.AzureFunctions/NetheriteProviderFactory.cs
sebastianburckhardt: Apparently it is not necessary for users to make a distinction of what order the code is running in. 

I stepped through the sample in the debugger to see why this is working...  what I observed is that the user code is running after me and calls AddSingleton, which I would have expected to cause some problem. Instead, everything ends up working as desired. The debugger shows that as a result, immediately after this step, there are two services of this type in the collection:
image
And when the NetheriteProviderFactory is constructed by DI, it is given the second one of these services, which is the desired result. 

It is a bit strange, but perhaps no need to worry since it is working correctly.

src/DurableTask.Netherite/Connections/ConnectionResolver.cs
sebastianburckhardt: I think I prefer the verbose name here. I find it helpful if implementations of an abstract class share the last word with the superclass.

src/DurableTask.Netherite/Connections/ConnectionResolver.cs
sebastianburckhardt: Yes, I suppose that makes sense.

src/DurableTask.Netherite/Connections/ConnectionResolver.cs
sebastianburckhardt: sure.

PR #200, opened 2022-11-15
Change MemoryF to SingleHost in samples

davidmrdavid:
As in title, to conform with new quickstart guidance


sebastianburckhardt:
This is waiting for the next NuGet release to be out (since the samples are using NuGet packages and the SingleHost was recentlly added).

PR #212, opened 2023-01-05
Add Functions isolated extension package

jviau:
This PR enables netherite usage in Azure functions dotnet isolated. 

- Adds a new project and nuget package `Microsoft.Azure.Functions.Worker.Extensions.DurableTask.Netherite`, which has a single attribute that tells the host to load the `Microsoft.Azure.DurableTask.Netherite.AzureFunctions` package.
    - Customers will reference this package in their dotnet isolated projects (and _not_ `Microsoft.Azure.DurableTask.Netherite.AzureFunctions`)
- Adds `dirs.proj` for release build, updates `azure-pipelines-release.yml` to use this.


src/Functions.Worker.Extensions.DurableTask.Netherite/Functions.Worker.Extensions.DurableTask.Netherite.csproj
```diff
@@ -0,0 +1,52 @@
...
+
+  <ItemGroup Condition="'$(Configuration)'=='Release'">
+    <Content Include="..\..\_manifest\**">
+      <Pack>true</Pack>
+      <PackagePath>content/SBOM</PackagePath>
+    </Content>
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Abstractions" Version="1.1.0" />
```
davidmrdavid, line 49: can we add a comment here to update the AssemblyInfo whenever we update this?

davidmrdavid (commented):
The change looks good to me, though I noticed the CI is failing due to the env variables not being read. Is that something you're looking into?

src/Functions.Worker.Extensions.DurableTask.Netherite/Functions.Worker.Extensions.DurableTask.Netherite.csproj
jviau: This is not the version that is to be kept in sync. It is the version of `Microsoft.Azure.DurableTask.Netherite.AzureFunctions`. Ultimately, it would be nice to automate this - but that can be left for a separate PR as we would need to extract some common versioning targets between the two projects.

jviau:
I have not looked too much into the CI failure, as I saw what you saw: a config issue, and I did not touch that project. So unsure what is going on. @sebastianburckhardt do you have context?

src/Functions.Worker.Extensions.DurableTask.Netherite/Functions.Worker.Extensions.DurableTask.Netherite.csproj
davidmrdavid: Ah, my mistake. I misread this (twice, one per PR). Thanks, this makes sense.

sebastianburckhardt (approved):
LGTM. The issue with the CI should resolve on its own once we merge this.

Issue #225, opened 2023-02-08
Durable Timers created with the fireAt set too far into the future causes an application crash


ksaunder:
I have a durable function that uses context.CreateTimer(utcFireAt, cts.Token) to implement a sort of cron job that will fire at a specific point in the future.  When I made the transition over to the Netherite backend I started to notice application crashes happening.  Upon further investigation it appears that the crash is happening within the ExpirationCheckLoop method in the BatchTimer class.
```
if (this.RequiresDelay(out var delay, out var due))
{
     var startWait = DateTime.UtcNow;

    
    this.notify.Wait(delay); // blocks thread until delay is over, or until notified
    this.tracer?.Invoke($"{this.name} is awakening at {(DateTime.UtcNow - due).TotalSeconds}s");
}
```
The issue is that the out var delay is calculated as a timespan of the difference between now and when the durable timer will fire in the future and the SemaphoreSlim class tries to turn that timespan into an int using the TotalMilliseconds and if it is larger than int.MaxValue it throws an ArgumentOutOfRangeException.  This exception ends up bringing down the whole process.

I have created a local alpha package with a modified copy of the method that instead sends in -1 as the delay if the delay timespan is larger than int.MaxValue and things appear to work.  Meaning the function app no longer crashes.
```
this.notify.Wait(delay.TotalMilliseconds > int.MaxValue ? -1 : (int)delay.TotalMilliseconds); 
```
However, I do not know quite enough about the BatchTimer class and what it is doing to know if modifying that Wait behavior to be indefinite is going to cause any long-term issues and it is a little hard to just test it since int.MaxValue milliseconds is a few weeks long.

sebastianburckhardt:
Yes, that is clearly a bug we want to fix ASAP. 

> However, I do not know quite enough about the BatchTimer class and what it is doing to know if modifying that Wait behavior to be indefinite is going to cause any long-term issues and it is a little hard to just test it since int.MaxValue milliseconds is a few weeks long.

I think what we want to do if the delay is too large is to just use some arbitrary shorter delay. This is safe because it is never a problem if the ExpirationCheckLoop runs sooner than necessary.

ksaunder:
Oh ok, actually my original fix before submitting this issue was to set the delay to int.MaxValue if the value was larger, but had the same questions about side effects.  I actually already have a local branch with that change.  I could easily turn that into a pull request and go from there.

sebastianburckhardt:
[...]

PR #226, opened 2023-02-09
Fix bug in BatchTimer

sebastianburckhardt:
As reported in #225, there is currently a bug in BatchTimer: if setting a timer too far in the future, the millisecond-delay exceeds int.MaxValue causing a crash.

The solution is to cap the delay at a safe value, here an arbitrarily chosen 1 day.  This does not interfere with the precision of the timer firing because the ExpirationCheckLoop after the delay passes, the code always rechecks to see if the timer is actually ready to fire based on the current time. It does mean the expiration check loop runs at least once a day, which is o.k.


davidmrdavid (approved):
LGTM

PR #227, opened 2023-02-23
Add FabricBot config for automatic GitHub triaging

davidmrdavid:
As in title. This will allow FabricBot to manage special labels like "needs: triage" and "needs: author response".

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -1,0 +1,536 @@
+{
+    "version": "1.0",
+    "tasks": [
+        {
+            "taskType": "trigger",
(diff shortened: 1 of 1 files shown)


sebastianburckhardt (approved):
I have no idea how it works but I will just trust you on this one.

davidmrdavid:
In fairness, this document is mostly autogenerated, so I don't think anyone except the FabricBot parser knows exactly how this works.

They are soon changing to a UI-based config model that should make our config much easier to maintain. Hoping that will remove the need for this file. In the meantime, thanks for the trust :)

Issue #229, opened 2023-03-01
Blob Storage Continually Growing using Netherite


UMCPGrad:
Our team is currently planning on switching to Netherite from the Azure Storage backend provider.  In testing, we see the improved latency, however we noticed that Blob Storage appears to continually grow over time.  We had a similar issue with the Azure Storage backend provider, and were hoping that we would not have to deal with clearing out blob storage.  Should we be running the Purge API daily to overcome this?  Here is a screen shot of about 24 hours worth of testing in our test environment.

<img width="1121" alt="Netherite" src="https://user-images.githubusercontent.com/17031822/222267040-17efc6cf-d539-4e71-bd2a-5da0d8e4bbec.png">

sebastianburckhardt:
Indeed. By default, an orchestration instance remains in storage indefinitely, until explicitly purged using the purge API). This behavior is the same across all storage providers. 

Supporting a more convenient auto-purge functionality is high on our priority list, but I cannot give you an ETA yet.

In the meantime, I would suggest to use a periodic timer function that purges completed orchestrations. Below are three examples (using a 10 minute, 1 hour, or 1 day interval) of how to run such a function. Which one of the three snippets you pick depends on how long you want to keep completed orchestrations in storage.

```csharp
        [FunctionName("PurgeEveryTenMinutes")]
        public static Task PurgeEveryTenMinutes(
            [DurableClient] IDurableOrchestrationClient client,
            [TimerTrigger("0 */10 * * * *")] TimerInfo myTimer)
        {
            // purge all orchestration instances that started at least 10 minutes ago
            // and are now in "Completed" state
            return client.PurgeInstanceHistoryAsync(
                DateTime.MinValue,
                DateTime.UtcNow.AddMinutes(-10),
                new List<OrchestrationStatus>
                {
                    OrchestrationStatus.Completed
                });
        }

        [FunctionName("PurgeEveryHour")]
        public static Task PurgeEveryHour(
             [DurableClient] IDurableOrchestrationClient client,
             [TimerTrigger("0 0 * * * *")] TimerInfo myTimer)
        {
            // purge all orchestration instances that started at least 1 hour ago
            // and are now in "Completed" state
            return client.PurgeInstanceHistoryAsync(
                DateTime.MinValue,
                DateTime.UtcNow.AddHours(-1),
                new List<OrchestrationStatus>
                {
                OrchestrationStatus.Completed
                });
        }

        [FunctionName("PurgeEveryDay")]
[...]

PR #230, opened 2023-03-07
Overhaul FASTER page and segment size parameters

sebastianburckhardt:
The following FASTER tuning parameters are immutable, i.e. must not be changed after a task hub is created (or else the storage provider cannot start any more):
- hybrid log page size
- hybrid log segment size
- event log page size
- event log segment size

However, sometimes we *do* need to change these parameters. Most recently (#229), we found that object logs are not collected aggressively enough after compaction, because they hybrid log segment size is much too large. Therefore, we want to change this default.

In this PR, we
- record the above sizes in the `taskhubparameters.json` file of the task hub, so that we can restore them when loading a task hub, which guarantees that parameters *always* match what is in storage even if the defaults change or the user explicitly changes them.
- change the default hybrid log page size to 512k to allow more aggressive collection of the object logs after compaction.

For compatibility, if loading a task hub that has no sizes recorded, we assume it uses the default sizes (as of the time of this PR).

### DIFF
--- a/src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
+++ b/src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
@@ -91,25 +91,26 @@ namespace DurableTask.Netherite.Faster
             public int? NumPagesToPreload;
         }
 
-        public FasterLogSettings GetDefaultEventLogSettings(bool useSeparatePageBlobStorage, FasterTuningParameters tuningParameters) => new FasterLogSettings
+        public FasterLogSettings GetEventLogSettings(bool useSeparatePageBlobStorage, FasterTuningParameters tuningParameters)
         {
-            LogDevice = this.EventLogDevice,
-            LogCommitManager = this.UseLocalFiles
-                ? null // TODO: fix this: new LocalLogCommitManager($"{this.LocalDirectoryPath}\\{this.PartitionFolderName}\\{CommitBlobName}")
-                : (ILogCommitManager)this,
-            PageSizeBits = tuningParameters?.EventLogPageSizeBits ?? 21, // 2MB
-            SegmentSizeBits = tuningParameters?.EventLogSegmentSizeBits ??
-                (useSeparatePageBlobStorage ? 35  // 32 GB
-                                            : 26), // 64 MB
-            MemorySizeBits = tuningParameters?.EventLogMemorySizeBits ?? 22, // 2MB
-        };
+            (int pageSizeBits, int segmentSizeBits) = BlobManager.GetImmutableEventLogParameters(useSeparatePageBlobStorage, tuningParameters);
 
-        public LogSettings GetDefaultStoreLogSettings(
-            bool useSeparatePageBlobStorage, 
-            long upperBoundOnAvailable, 
-            FasterTuningParameters tuningParameters)
+            return new FasterLogSettings
+            {
+                PageSizeBits = pageSizeBits,
+                SegmentSizeBits = segmentSizeBits,
+                 LogDevice = this.EventLogDevice,
+                LogCommitManager = this.UseLocalFiles
+                    ? null // TODO: fix this: new LocalLogCommitManager($"{this.LocalDirectoryPath}\\{this.PartitionFolderName}\\{CommitBlobName}")
+                    : (ILogCommitManager)this,
+                MemorySizeBits = tuningParameters?.EventLogMemorySizeBits ?? 22, // 2MB
+            };
+        }
+
+        public LogSettings GetStoreLogSettings(bool useSeparatePageBlobStorage, long upperBoundOnAvailable, FasterTuningParameters tuningParameters) 
         {
-            int pageSizeBits = tuningParameters?.StoreLogPageSizeBits ?? 10; // default page size is 1k
+            // start with page and segment sizes
+            (int pageSizeBits, int segmentSizeBits) = BlobManager.GetImmutableStoreLogParameters(useSeparatePageBlobStorage, tuningParameters);
 
             // compute a reasonable memory size for the log considering maximally available memory, and expansion factor
             int memorybits = 0;
@@ -134,11 +135,9 @@ namespace DurableTask.Netherite.Faster
             {
                 LogDevice = this.HybridLogDevice,
                 ObjectLogDevice = this.ObjectLogDevice,
-                PageSizeBits = pageSizeBits,  
+                PageSizeBits = pageSizeBits,
                 MutableFraction = tuningParameters?.StoreLogMutableFraction ?? 0.9,
-                SegmentSizeBits = tuningParameters?.StoreLogSegmentSizeBits ??
-                    (useSeparatePageBlobStorage ? 35   // 32 GB
-                                                : 32), // 4 GB
+                SegmentSizeBits = segmentSizeBits,
                 PreallocateLog = false,
                 ReadFlags = ReadFlags.None,
                 ReadCacheSettings = null, // no read cache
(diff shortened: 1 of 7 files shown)


src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
```diff
@@ -91,25 +91,26 @@ public class FasterTuningParameters
...
-            long upperBoundOnAvailable, 
-            FasterTuningParameters tuningParameters)
+            return new FasterLogSettings
+            {
+                PageSizeBits = pageSizeBits,
+                SegmentSizeBits = segmentSizeBits,
+                 LogDevice = this.EventLogDevice,
+                LogCommitManager = this.UseLocalFiles
+                    ? null // TODO: fix this: new LocalLogCommitManager($"{this.LocalDirectoryPath}\\{this.PartitionFolderName}\\{CommitBlobName}")
+                    : (ILogCommitManager)this,
```
davidmrdavid, line 105: can you please elaborate on this TODO comment? What needs fixing exactly?

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
```diff
@@ -134,18 +135,34 @@ public class FasterTuningParameters
...
+
+            return (pageSizeBits, segmentSizeBits);
+        }
+
+        public static (int pageSizeBits, int segmentSizeBits) GetImmutableStoreLogParameters(bool useSeparatePageBlobStorage, FasterTuningParameters tuningParameters)
+        {
+            int pageSizeBits =    tuningParameters?.StoreLogPageSizeBits    ?? 10; // 1kB
+            int segmentSizeBits = tuningParameters?.StoreLogSegmentSizeBits ?? 19; // 512 kB
+             
+            return (pageSizeBits, segmentSizeBits);
```
davidmrdavid, line 164: It appears the first parameter is not being used here, or is it?

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
```diff
@@ -134,18 +135,34 @@ public class FasterTuningParameters
...
                 PreallocateLog = false,
                 ReadFlags = ReadFlags.None,
                 ReadCacheSettings = null, // no read cache
                 MemorySizeBits = memorybits,
             };
         }
 
+        public static (int pageSizeBits, int segmentSizeBits) GetImmutableEventLogParameters(bool useSeparatePageBlobStorage, FasterTuningParameters tuningParameters)
+        {
+            int pageSizeBits = tuningParameters?.EventLogPageSizeBits ?? 21; // 2MB
```
davidmrdavid, line 150: My bit-arithmetic fails me here. How are you going from 21 bits (?) to 2MB?

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
```diff
@@ -157,10 +174,21 @@ public class FasterTuningParameters
 
         public static string GetStorageFormat(NetheriteOrchestrationServiceSettings settings)
```
davidmrdavid, line 175: I noticed this doesn't seem to be called in this PR. Is it called elsewhere?

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
```diff
@@ -201,6 +232,28 @@ public static void CheckStorageFormat(string format, NetheriteOrchestrationServi
                 {
                     throw new NetheriteConfigurationException($"The current storage format version (={StorageFormatVersion.Last()}) is incompatible with the existing taskhub (={taskhubFormat.FormatVersion}).");
                 }
+
+                // read the immutable log parameters from storage, and keep them as a tuning parameter
+                // (which may override any tuning parameters set by the user; that is what we want since they cannot be honored)
```
davidmrdavid, line 237: can we throw a warning if we detect this conflict: that the new parameters cannot be honored?

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
```diff
@@ -187,7 +218,7 @@ class StorageFormatSettings
             Formatting = Formatting.None,
         };
 
-        public static void CheckStorageFormat(string format, NetheriteOrchestrationServiceSettings settings)
+        public static void CheckAndLoadStorageFormat(string format, NetheriteOrchestrationServiceSettings settings)
```
davidmrdavid, line 221: super nitpick: at first glance, I find it a bit unintuitive to first _check_ and then _load_ the config here. The opposite order makes more sense to me, as it allows us to validate the config we just loaded. I realize this current ordering works because we don't really throw any validation errors for the config we _load_.

I'll make a _soft_ suggestion to rename this to `LoadAndCheckStorageFormat` and check the order of operations to match that new name, which I think is more future-proof. But feel free to disregard this comment as it's mostly a stylistic nit in the present.

davidmrdavid (commented):
Left some comments!

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
sebastianburckhardt: right, this was just a copy-paste (so it has nothing to do with this PR). The TODO here refers to the fact that at some point Netherite supported using the file system (instead of Azure Storage) to store task hubs. But it is not an official feature and commented out at the moment. There is some scenario in which we bring this back on K8s, therefore I did not want to just delete it outright.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
sebastianburckhardt: I will remove it. It was used in a previous version.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
sebastianburckhardt: 2^21 bytes = (2^1)(2^10)(2^10) bytes = 2 * 1024 * 1024 bytes = 2 MB

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
sebastianburckhardt: Yes, it is called when a new taskhub is created in `IStorageLayer.CreateTaskhubIfNotExistsAsync`. At that point, all the important parameters for this task hub storage format are recorded inside `taskhubparameters.json`.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
sebastianburckhardt: That is a good idea. I think we used to have a warning like that for the partition count (which is also an immutable parameter) but I can't find it anymore in the code. It should be there also.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
davidmrdavid: ah gotcha. That makes sense.
Might be helpful just to add a blurb to that comment that this local-files support is for K8s but not advertised today.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
davidmrdavid: I should have picked up on this 😅  Thanks!

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
sebastianburckhardt: I have reworked this a bit to make it clearer.

davidmrdavid (approved):
LGTM! Thanks for these changes

Issue #231, opened 2023-03-14
Testing with 100 orchestrators and 1000 activities each, getting this exception: Part10 !!! Assertion Failed: bad fragment sequence


ericleigh007:
i'm doing some load testing comparing Netherite with the default back end and I must say this backend is so much better in terms of latencies and throughput.

Great job.  We are very much looking forward to amping up our Durable Functions performance now that this is GA.

however, I'm getting an error every once in a while when doing some maximum test, in this case 200 orchestrators that each launch off 1000 activities.

The error looks like this
````
2023-03-14T00:32:11Z   [Error]   Part10 !!! Assertion Failed: bad fragment sequence: position in    at System.Environment.get_StackTrace()
   at DurableTask.Netherite.Partition.Assert(Boolean condition, String message) in /_/src/DurableTask.Netherite/OrchestrationService/Partition.cs:line 164
   at DurableTask.Netherite.FragmentationAndReassembly.Reassemble[TEvent](IEnumerable`1 earlierFragments, IEventFragment lastFragment, Partition partition) in /_/src/DurableTask.Netherite/Util/FragmentationAndReassembly.cs:line 86
   at DurableTask.Netherite.ReassemblyState.Process(PartitionEventFragment evt, EffectTracker effects) in /_/src/DurableTask.Netherite/PartitionState/ReassemblyState.cs:line 61
   at DurableTask.Netherite.EffectTracker.ProcessEffectOn(TrackedObject trackedObject) in /_/src/DurableTask.Netherite/Abstractions/PartitionState/EffectTracker.cs:line 80
   at DurableTask.Netherite.Faster.FasterKV.ProcessEffectOnTrackedObject(Key k, EffectTracker tracker) in /_/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs:line 785
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
   at DurableTask.Netherite.Faster.FasterKV.ProcessEffectOnTrackedObject(Key k, EffectTracker tracker)
   at DurableTask.Netherite.Faster.TrackedObjectStoreEffectTracker.ApplyToStore(TrackedObjectKey key, EffectTracker tracker) in /_/src/DurableTask.Netherite/StorageLayer/Faster/TrackedObjectStoreEffectTracker.cs:line 26
[...]

PR #233, opened 2023-03-15
Fix fragmentation/reassembly logic

sebastianburckhardt:
As discovered in #231, there is a bug in the fragmentation/reassembly logic that is exposed in situations where a partition is moved to a different node while it is sending a fragmented event. In that case, fragments of the old sequence can interleave with fragments of the new sequence, which breaks the reassembly logic. 

To fix this, this PR changes the logic so it uses a unique identifier Guid, instead of the event id, to track groups of fragments that belong together. This guarantees that even in interleaving situations, we correctly track each sequence separately.

For compatibility, this change adds the Guid as a nullable new field to the fragment. Old fragments can thus still be deserialized as before, and are treated the same way as before (which is important for not breaking existing task hubs since the event sourcing may replay events with the expectation of deterministic results).

### DIFF
--- a/src/DurableTask.Netherite/OrchestrationService/Client.cs
+++ b/src/DurableTask.Netherite/OrchestrationService/Client.cs
@@ -87,10 +87,9 @@ namespace DurableTask.Netherite
             if (clientEvent is ClientEventFragment fragment)
             {
                 var originalEventString = fragment.OriginalEventId.ToString();
-
                 var group = fragment.GroupId.HasValue
                   ? fragment.GroupId.Value.ToString()       // groups are now the way we track fragments
-                  : $"{fragment.OriginalEventId}-{fragment.ReceiveChannel}";  // prior to introducing groups, we used event id and channel, which is not always good enough
+                  : $"{originalEventString}-{fragment.ReceiveChannel}";  // prior to introducing groups, we used event id and channel, which is not always good enough
 
                 if (this.traceHelper.LogLevelLimit == Microsoft.Extensions.Logging.LogLevel.Trace)
                 {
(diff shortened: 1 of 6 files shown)


src/DurableTask.Netherite/OrchestrationService/Client.cs
```diff
@@ -87,7 +87,10 @@ public void Process(ClientEvent clientEvent)
             if (clientEvent is ClientEventFragment fragment)
             {
                 var originalEventString = fragment.OriginalEventId.ToString();
-                var index = (originalEventString, clientEvent.ReceiveChannel);
+
+                var group = fragment.GroupId.HasValue
+                  ? fragment.GroupId.Value.ToString()       // groups are now the way we track fragments
+                  : $"{fragment.OriginalEventId}-{fragment.ReceiveChannel}";  // prior to introducing groups, we used event id and channel, which is not always good enough
```
cgillum, line 93: Do you still need `originalEventString` on line 89? I was expecting that you'd reuse that local variable here but maybe it can be removed?

src/DurableTask.Netherite/OrchestrationService/Client.cs
```diff
@@ -105,6 +108,8 @@ public void Process(ClientEvent clientEvent)
 
                     var reassembledEvent = FragmentationAndReassembly.Reassemble<ClientEvent>(stream, fragment);
 
+                    this.Fragments.Remove(group);
```
cgillum, line 111: This appears to be new behavior. Is this part of the bug fix?

sebastianburckhardt: I moved this here from line 136 (which was executed by the call on the next line). It should be logically equivalent.

src/DurableTask.Netherite/OrchestrationService/Client.cs
sebastianburckhardt: Good catch. Yes, it can be removed.

sebastianburckhardt: Actually, I just realized it is being used by trace statements further down. So I will reuse it.

cgillum (approved):

PR #242, opened 2023-04-04
Add semaphore example to PerformanceBenchmarks

sebastianburckhardt:
This example demonstrates how one can use an entity to implement a semaphore that acts as a global concurrency-limiter for activities.

This was motivated by the discussion in https://github.com/Azure/azure-functions-durable-extension/issues/2434.

To run the tests locally,

1. Modify host.json in the PerfomanceTests project if you want to use AzureStorage provider instead of Netherite
2.  Set environment variables for `AzureWebJobsStorage` and `EventHubsConnection` (don't need the latter unless using Netherite)
3. Start PerformanceTests function app 
4. Launch 1000 orchestrations by `curl http://localhost:7071/start -d OrchestrationWithSemaphore.1000`
5. Observe progress of the orchestrations by calling `curl http://localhost:7071/query`
6. Observe state of the semaphore by calling  `curl http://localhost:7071/semaphore`

FYI, when I ran quick tests on my local machine, Netherite completed all the orchestrations in 38 seconds and Azure Storage backend completed them in 190 seconds.

### DIFF
--- a/test/PerformanceTests/Benchmarks/Semaphore/HttpTriggers.cs
+++ b/test/PerformanceTests/Benchmarks/Semaphore/HttpTriggers.cs
@@ -1,0 +1,49 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace PerformanceTests.HelloCities
+{
+    using System;
+    using System.IO;
+    using System.Threading.Tasks;
+    using Microsoft.AspNetCore.Mvc;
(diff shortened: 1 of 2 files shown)


sebastianburckhardt:
> What do you think about mirroring this in the extension repo? Seems useful

yes - I can see using this as a reference. But it is easier to just point them to this PR in this repo I think, given that it already has setup instructions and can run both Netherite and Azure Storage provider.

PR #248, opened 2023-04-12
Fix null metrics exceptions

### LINKED ISSUE #246: Fix ArgumentNullException in ScaleMonitor
It appears that the scale monitor is sometimes called with null metrics and throws exceptions. 
It would be preferrable to handle this gracefully.

Example of the reported excception:
image

sebastianburckhardt:
Sometimes the scale monitor is called with null metrics. These are transient occurrences. Instead of throwing an exception, we handle this gracefully.

Fixes #246.

### DIFF
--- a/src/DurableTask.Netherite.AzureFunctions/NetheriteProvider.cs
+++ b/src/DurableTask.Netherite.AzureFunctions/NetheriteProvider.cs
@@ -205,14 +205,16 @@ namespace DurableTask.Netherite.AzureFunctions
             {
                 ScaleRecommendation recommendation;               
                 try
-                { 
-                    if (metrics == null || metrics.Length == 0)
+                {
+                    var lastMetric = metrics?.LastOrDefault(m => m?.Metrics != null);
+
+                    if (lastMetric == null)
                     {
                         recommendation = new ScaleRecommendation(ScaleAction.None, keepWorkersAlive: true, reason: "missing metrics");
                     }
                     else
                     {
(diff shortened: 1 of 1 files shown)


src/DurableTask.Netherite.AzureFunctions/NetheriteProvider.cs
```diff
@@ -205,14 +205,16 @@ ScaleStatus GetScaleStatusCore(int workerCount, NetheriteScaleMetrics[] metrics)
             {
                 ScaleRecommendation recommendation;               
                 try
-                { 
-                    if (metrics == null || metrics.Length == 0)
```
davidmrdavid, line 209: is this expected behavior, or just something that we know is happening transiently and that we do not want to fail catastrophically?

If it's expected, then LGTM (and will approve as such). However, if it's not expected, should we log something?

davidmrdavid (approved):

src/DurableTask.Netherite.AzureFunctions/NetheriteProvider.cs
sebastianburckhardt: I am not aware of any specification for the scale controller API so I can't tell whether it is expected or not. But it does happen transiently.

As with other scale controller abnormalities, it is not clear to me what would be the best way to log/report them. We are already logging the "missing metrics" description as part of the scale decision, so it is not completely invisible. We could also generate a warning.

Issue #245, opened 2023-04-10
Fix event hubs "quota exceeded" exceptions


sebastianburckhardt:
In some cases Netherite users have seen exceptions like the following:

```Microsoft.Azure.EventHubs.QuotaExceededException: Exceeded the maximum number of allowed receivers per partition in a consumer group which is 5.```

There appear to be multiple reasons that contribute to this. Both of these issues should be investigated and fixed.

- The client hash algorithm does not retry, nor generate a useful error message, when there are more than 5 clients in a bucket, which can happen. I saw in traces that even with 128 hash buckets, it may take only 256 elements to cause exhaustion due to normal imbalance. Thus we need to deal with this case better. For example, we can change the client id and retry.

- EH clients are shutdown late in the shutdown process, which appears to sometimes hang, causing clients to not be released. Thus the number of "live" clients can be much larger than the number of nodes in the system. We should (a) shut down clients sooner, and (b) investigate and fix the underlying causes for the hangs.

PR #249, opened 2023-04-12
Optimize host shutdown

sebastianburckhardt:
Currently, when the EventHubsTransport shuts down, it shuts down the client and closes the EH connection *only after* all the workers are shut down.

This can cause problems if the worker shut down hangs as observed in #245, because the EH client does not shut down, causing issues.

This PR modifies the shut down so that clients & connections are shut down in parallel with the workers. This fixes the problem described above, and also improves the speed of the shutdown process.

### DIFF
--- a/src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsTransport.cs
+++ b/src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsTransport.cs
@@ -159,7 +159,7 @@ namespace DurableTask.Netherite.EventHubsTransport
             {
                 if (this.settings.PartitionManagement != PartitionManagementOptions.Scripted)
                 {
-                    this.traceHelper.LogInformation($"Registering Partition Host with EventHubs");
+                    this.traceHelper.LogInformation($"EventHubsTransport is registering PartitionHost");
 
                     string formattedCreationDate = this.connections.CreationTimestamp.ToString("o").Replace("/", "-");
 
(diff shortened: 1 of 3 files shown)


src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsTransport.cs
```diff
@@ -289,39 +289,53 @@ public IEventProcessor CreateEventProcessor(PartitionContext context)
...
 
-            this.traceHelper.LogDebug("Stopping client");
-            await this.client.StopAsync();
+        async Task StopWorkersAsync()
+        {
+            this.traceHelper.LogDebug("EventHubsTransport is stopping partition and loadmonitor hosts");
+            await Task.WhenAll(
+              this.StopPartitionHostAsync(),
+              this.StopLoadMonitorHostAsync());
+        }
```
davidmrdavid, line 308: Do we want a log indicating that the WhenAll completed?

src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsTransport.cs
```diff
@@ -289,39 +289,53 @@ public IEventProcessor CreateEventProcessor(PartitionContext context)
...
+            this.traceHelper.LogDebug("EventHubsTransport is stopping client process loop");
+            await this.clientProcessTask;
+
+            this.traceHelper.LogDebug("EventHubsTransport is closing connections");
             await this.connections.StopAsync();
 
-            this.traceHelper.LogInformation("EventHubsBackend shutdown completed");
+            this.traceHelper.LogDebug("EventHubsTransport is stopping client");
+            await this.client.StopAsync();
         }
```
davidmrdavid, line 320: do we want a log indicating the client _stopped_, not just that it began its stop operation?

davidmrdavid (approved):
LGTM - just two suggestions for extra logging. Please consider adding them if you think the granularity will help with debugging.

src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsTransport.cs
sebastianburckhardt: yes, that makes sense.

sebastianburckhardt:
The CI failures are likely not related. I have seen them a few times already; will work on that separately.

PR #250, opened 2023-04-12
Fix race condition in OutboxState

sebastianburckhardt:
Fixes a problem discovered during testing.

While iterating over the collection of messages to send, even if this is just a single message, the send can actually complete, and the send confirmation can also complete, at which point the element may be removed from the collection, which causes a conflict with the initial loop iteration as we see here:
```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at DurableTask.Netherite.OutboxState.Send(Batch batch) in D:\git\durabletask-netherite\src\DurableTask.Netherite\PartitionState\OutboxState.cs:line 112
   at DurableTask.Netherite.DurabilityListeners.ConfirmDurable(Event evt) in D:\git\durabletask-netherite\src\DurableTask.Netherite\Util\DurabilityListeners.cs:line 79
   at DurableTask.Netherite.Faster.LogWorker.Process(IList`1 batch) in D:\git\durabletask-netherite\src\DurableTask.Netherite\StorageLayer\Faster\LogWorker.cs:line 168
```

This PR fixes the problem by copying the collection into an array before iterating over it.

### DIFF
--- a/src/DurableTask.Netherite/PartitionState/OutboxState.cs
+++ b/src/DurableTask.Netherite/PartitionState/OutboxState.cs
@@ -108,7 +108,8 @@ namespace DurableTask.Netherite
 
             if (batch.OutgoingResponses.Count > 0)
             {
-                foreach (var outresponse in batch.OutgoingResponses.ToArray())
+                var outgoingResponses = batch.OutgoingResponses.ToArray(); //copy for safe iteration
+                foreach (var outresponse in outgoingResponses )
                 {
                     DurabilityListeners.Register(outresponse, batch);
                     this.Partition.Send(outresponse);
@@ -116,7 +117,8 @@ namespace DurableTask.Netherite
             }
             if (batch.OutgoingMessages.Count > 0)
             {
-                foreach (var outmessage in batch.OutgoingMessages.ToArray())
+                var outgoingMessages =  batch.OutgoingMessages.ToArray(); // copy for safe iteration
+                foreach (var outmessage in outgoingMessages)
                 {
                     DurabilityListeners.Register(outmessage, batch);
                     this.Partition.Send(outmessage);
--- a/src/DurableTask.Netherite/PartitionState/OutboxState.cs
+++ b/src/DurableTask.Netherite/PartitionState/OutboxState.cs
@@ -99,20 +99,28 @@ namespace DurableTask.Netherite
         void Send(Batch batch)
         {
             batch.ReadyToSendTimestamp = this.Partition.CurrentTimeMs;
-            var outMessages = batch.OutgoingMessages.Count < 2 ? batch.OutgoingMessages : batch.OutgoingMessages.ToList();// prevent concurrent mod
-            batch.TotalAcksExpected = batch.OutgoingResponses.Count + outMessages.Count;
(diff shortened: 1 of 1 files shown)


src/DurableTask.Netherite/PartitionState/OutboxState.cs
```diff
@@ -99,20 +99,28 @@ public void ConfirmDurable(Event evt)
...
+
+            if (batch.OutgoingResponses.Count > 0)
             {
-                DurabilityListeners.Register(outresponse, batch);
-                this.Partition.Send(outresponse);
+                foreach (var outresponse in batch.OutgoingResponses.ToArray())
+                {
+                    DurabilityListeners.Register(outresponse, batch);
+                    this.Partition.Send(outresponse);
+                }
```
davidmrdavid, line 115: Am I right in understanding that you're copying the responses _inline_ by calling `.ToArray()`?
If so - I think it would be better to split the copying of the collection into a separate line: otherwise, I think it's easy to miss that the `ToArray()` step is important for correctness. Please consider this:

```suggestion
                var outgoingResponses = batch.OutgoingResponses.ToArray(); //copy for safe iteration
                foreach (var outresponse in outgoingResponses )
                {
                    DurabilityListeners.Register(outresponse, batch);
                    this.Partition.Send(outresponse);
                }
```

src/DurableTask.Netherite/PartitionState/OutboxState.cs
```diff
@@ -99,20 +99,28 @@ public void ConfirmDurable(Event evt)
...
-            foreach (var outmessage in outMessages)
+            if (batch.OutgoingMessages.Count > 0)
             {
-                DurabilityListeners.Register(outmessage, batch);
-                this.Partition.Send(outmessage);
+                foreach (var outmessage in batch.OutgoingMessages.ToArray())
+                {
+                    DurabilityListeners.Register(outmessage, batch);
+                    this.Partition.Send(outmessage);
+                }
```
davidmrdavid, line 123:
```suggestion
                var outgoingMessages =  batch.OutgoingMessages.ToArray(); // copy for safe iteration
                foreach (var outmessage in outgoingMessages)
                {
                    DurabilityListeners.Register(outmessage, batch);
                    this.Partition.Send(outmessage);
                }
```

davidmrdavid (commented):
Some nits and questions

src/DurableTask.Netherite/PartitionState/OutboxState.cs
```diff
@@ -99,20 +99,28 @@ public void ConfirmDurable(Event evt)
         void Send(Batch batch)
         {
             batch.ReadyToSendTimestamp = this.Partition.CurrentTimeMs;
-            var outMessages = batch.OutgoingMessages.Count < 2 ? batch.OutgoingMessages : batch.OutgoingMessages.ToList();// prevent concurrent mod
-            batch.TotalAcksExpected = batch.OutgoingResponses.Count + outMessages.Count;
```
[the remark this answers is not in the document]
sebastianburckhardt: This logic was based on the assumption that concurrent modification is only a problem if the list is larger than one element. That assumption was wrong. So we now *always* make a copy of the list.

sebastianburckhardt:
Thanks for the review. I have committed your suggestions.

Issue #236, opened 2023-03-17
InstanceCount reported in DurableTaskPartitions table is not accurate


sebastianburckhardt:
Seeing executions where the reported instance count is negative in the DurableTaskPartitions table:

![image](https://user-images.githubusercontent.com/6415760/225966469-da322ec8-ad3b-4e07-9ef4-b00fe94f8e36.png)

This means that the logic that maintains the instance count within each partition is somehow broken, thus we cannot trust the reported instance count.

sebastianburckhardt:
Found more serious consequences of this, likely caused by the same problem:
![image](https://user-images.githubusercontent.com/6415760/231290998-16171fa7-49c7-4074-b53b-ba7aec38a370.png)

sebastianburckhardt:
After running more long-duration tests I was able to repro this and found that the likely culprit is that occasionally, FASTER may execute an RMW operation more than once (even if the operation succeeds). This is not wrong in the sense that the Functions API does not say it won't ever do that. However, in my object representation (where I apply an update event to a tracked object) an RMW operation has side effects (it does not just modify the state of the value for the key but queues other locations that should be updated) and those side effects must not be duplicated. 

I will fix this by adding deduplication check which prevents the same update event from updating the same object twice.

PR #252, opened 2023-04-20
Fix double application of partition events on tracked objects

sebastianburckhardt:
fix bug where an effect is applied twice due to an RMW executing twice in FASTER, as discovered and discussed in #236. The results are pretty unpredictable, state can get corrupted.

The fix is to record the log commit number of the last-applied event in the tracked object, and then use that to filter duplicate applications of the same event. It is not necessary to store this number with the tracked object, in-memory-only is sufficient (since the double RMW does not actually get committed).

This may also have had other manifestations that went undetected.

### DIFF
--- a/src/DurableTask.Netherite/Abstractions/PartitionState/EffectTracker.cs
+++ b/src/DurableTask.Netherite/Abstractions/PartitionState/EffectTracker.cs
@@ -77,8 +77,12 @@ namespace DurableTask.Netherite
         {
             try
             {
-                this.currentUpdate.ApplyTo(trackedObject, this);
-                trackedObject.Version++;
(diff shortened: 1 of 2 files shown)


davidmrdavid (approved):
LGTM but curious about the following:
You mention the duplicate RMW effect isn't stricly forbidden, but is it explictly allowed? Wondering if this behavior is something the FASTER folks are aware of.

Also - are there other places where we utilize RMW operations in Netherite that may need to be updated with a similar mitigation?

sebastianburckhardt:
> but is it explictly allowed

I think I am using FASTER in a weird way - I am overriding some default behaviors. Normal users should not notice this problem.

> Also - are there other places where we utilize RMW operations in Netherite that may need to be updated with a similar mitigation?

Luckily, no. This is the one and only place that updates tracked objects.

PR #253, opened 2023-04-20
Fix inconsistent handling of expiration for partition event fragments

sebastianburckhardt:
In my testing I noticed that defragmentation of events was failing occasionally.

After some analysis I found that the existing code was handling the expiration of fragments inconsistently:
- during recovery, fragments deemed irrelevant (because they belong to already-timed out client requests or already-processed messages from other partitions) were removed from the `ReassemblyState` tracked object. This is a correct idea that is necessary to avoid keeping such fragments forever in circumstances where the group will never be completed.
- but incoming fragments were still always looking up previous fragments in the group, causing crashes because what they were looking for had already been removed.

This PR fixes these crashes by remembering the horizon of what it has removed; it can the throw away incoming fragments whose group was already removed immediately instead of looking them up.


sebastianburckhardt:
The complexity of this solution is not worth the benefits. A much simpler solution would be to just ignore fragments if their group is missing. I will close this PR and create a new one.

PR #254, opened 2023-04-20
Fix data race in memory tracker

sebastianburckhardt:
Can cause divide-by-zero exceptions. It happens quite rarely. 

I observed this during long-running testing in the consumption plan.

### DIFF
--- a/src/DurableTask.Netherite/StorageLayer/Faster/MemoryTracker.cs
+++ b/src/DurableTask.Netherite/StorageLayer/Faster/MemoryTracker.cs
@@ -41,7 +41,8 @@ namespace DurableTask.Netherite.Faster
 
         public void UpdateTargetSizes()
         {
-            if (this.stores.Count > 0)
+            int numberOfStores = this.stores.Count;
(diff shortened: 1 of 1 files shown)


src/DurableTask.Netherite/StorageLayer/Faster/MemoryTracker.cs
```diff
@@ -41,7 +41,8 @@ public CacheTracker NewCacheTracker(FasterKV store, int partitionId, CacheDebugg
 
         public void UpdateTargetSizes()
         {
-            if (this.stores.Count > 0)
+            int numberOfStores = this.stores.Count;
+            if (numberOfStores > 0)
```
davidmrdavid, line 45: nit: let's document the race condition, since it's non-obious
```suggestion
            // store count to avoid race condition where count goes to zero and causes divide-by-zero error
            int numberOfStores = this.stores.Count;
            if (numberOfStores > 0)
```

davidmrdavid (approved):
That's quite the race condition! Worth documenting it imo.

PR #258, opened 2023-04-21
Improve trace messages for non-executable instances

sebastianburckhardt:
Makes several changes to how non-executable instances create warnings, to help users interpret them:
- distinguishes between entities and orchestrations in the error message
- uses clearer message text for describing the situation
- does not create any warnings for the common case where a TimerFiredEvent targets a completed orchestration, because this happens commonly for canceled timers

### DIFF
--- a/src/DurableTask.Netherite/OrchestrationService/OrchestrationMessageBatch.cs
+++ b/src/DurableTask.Netherite/OrchestrationService/OrchestrationMessageBatch.cs
@@ -118,16 +118,22 @@ namespace DurableTask.Netherite
                 this.workItem.EventCount = historyState.History?.Count ?? 0;
             }
 
-            if (!this.IsExecutableInstance(this.workItem, out var reason))
+            if (!this.IsExecutableInstance(this.workItem, out var reason, out bool traceWarningEvenForTimers))
             {
                 // this instance cannot be executed, so we are discarding the messages
                 for (int i = 0; i < this.workItem.MessageBatch.MessagesToProcess.Count; i++)
                 {
-                    partition.WorkItemTraceHelper.TraceTaskMessageDiscarded(
-                        this.PartitionId,
-                        this.workItem.MessageBatch.MessagesToProcess[i],
-                        this.workItem.MessageBatch.MessagesToProcessOrigin[i],
-                        reason);
+                    TaskMessage message = this.workItem.MessageBatch.MessagesToProcess[i];
+
+                    // since this can happen by design for canceled timers, we are not always creating warnings
+                    if (traceWarningEvenForTimers || ! (message.Event is TimerFiredEvent))
+                    {
+                        partition.WorkItemTraceHelper.TraceTaskMessageDiscarded(
+                            this.PartitionId,
+                            message,
+                            this.workItem.MessageBatch.MessagesToProcessOrigin[i],
+                            reason);
+                    }
                 }
 
                 // we issue a batch processed event which will remove the messages without updating the instance state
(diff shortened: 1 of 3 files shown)


src/DurableTask.Netherite/Tracing/ClientTraceHelper.cs
```diff
@@ -111,11 +111,11 @@ public void TraceSend(PartitionEvent @event)
             {
                 if (this.logger.IsEnabled(LogLevel.Debug))
                 {
-                    this.logger.LogDebug("{client} Sending event {eventId}: {event}", this.tracePrefix, @event.EventIdString, @event);
+                    this.logger.LogDebug("{client} Sending event {eventId} to partition {partitionId}: {event}", this.tracePrefix, @event.EventIdString, @event.PartitionId, @event);
```
davidmrdavid, line 114: just for my .NET knowledge: what does the `@` in `@event.PartitionId` do?

src/DurableTask.Netherite/OrchestrationService/OrchestrationMessageBatch.cs
```diff
@@ -162,20 +168,36 @@ public void ConfirmDurable(Event evt)
...
+                    }
+                    else if (workItem.OrchestrationRuntimeState.Events.Count == 0)
+                    {
+                        message = "Instance does not exist (it may have been purged)";
+                        traceWarningEvenForTimers = false;
+                    }
+                    else
+                    {
+                        message = $"Instance has an invalid history ({workItem.OrchestrationRuntimeState.Events} events)";
+                        traceWarningEvenForTimers = true;
```
davidmrdavid, line 198: I'm not immediately seeing the relationship between these if-else cases, and the different values for `traceWarningEvenForTimers`. Can you please clarify them?

src/DurableTask.Netherite/OrchestrationService/OrchestrationMessageBatch.cs
```diff
@@ -118,16 +118,22 @@ public override void OnReadComplete(TrackedObject s, Partition partition)
                 this.workItem.EventCount = historyState.History?.Count ?? 0;
             }
 
-            if (!this.IsExecutableInstance(this.workItem, out var reason))
+            if (!this.IsExecutableInstance(this.workItem, out var reason, out bool traceWarningEvenForTimers))
```
davidmrdavid, line 121: nitpick: would it be better to just call this `traceTimerWarnings`?

davidmrdavid (commented):
Some questions

src/DurableTask.Netherite/Tracing/ClientTraceHelper.cs
sebastianburckhardt: Because `event` is a keyword in C#, you can't use it unless you prefix it `@`. You could say the '@' is an un-keyword operator.

I actually don't use this very often anymore because it is a bit distracting. I think it is usually preferrable to just use a different name.

src/DurableTask.Netherite/OrchestrationService/OrchestrationMessageBatch.cs
sebastianburckhardt: I can add comments. The point is that some of these warnings are common place when using timers. I have tests where timers are canceled, the orchestration completes, and then the orchestration is purged. But when the timer still fires later, we get a message for a non-existent orchestration. Because this is completely within the normal operation and use of orchestrations, I do not want to issue a warning in this case.

src/DurableTask.Netherite/OrchestrationService/OrchestrationMessageBatch.cs
sebastianburckhardt: I can see that this is confusing, because of the way the logic is stuctured. I will revise the logic so the decisions are all made inside `this.IsExecutableInstance`.

src/DurableTask.Netherite/Tracing/ClientTraceHelper.cs
davidmrdavid: oh, fancy. Thanks for clarifying

src/DurableTask.Netherite/OrchestrationService/OrchestrationMessageBatch.cs
```diff
@@ -162,20 +169,36 @@ public void ConfirmDurable(Event evt)
...
         {
+            message = default;
+
             if (workItem.OrchestrationRuntimeState.ExecutionStartedEvent == null
                 && !this.MessagesToProcess.Any(msg => msg.Event is ExecutionStartedEvent))
             {
                 if (DurableTask.Core.Common.Entities.AutoStart(this.InstanceId, this.MessagesToProcess))
                 {
-                    message = default;
+                    isNormalSituationForTimerFired = default;
```
davidmrdavid, line 181: nit: for booleans, I'd explicitly specify the default. I think "default" is best used for values that don't directly affect control-flow like booleans. That said, I realize in this case it's pretty obvious the default is `false` so I'll leave this change up to you.

src/DurableTask.Netherite/OrchestrationService/OrchestrationMessageBatch.cs
```diff
@@ -184,10 +207,11 @@ bool IsExecutableInstance(TaskOrchestrationWorkItem workItem, out string message
                 workItem.OrchestrationRuntimeState.OrchestrationStatus != OrchestrationStatus.Pending)
             {
                 message = $"Instance is {workItem.OrchestrationRuntimeState.OrchestrationStatus}";
+                isNormalSituationForTimerFired = true;
                 return false;
             }
 
-            message = null;
+            isNormalSituationForTimerFired = default;
```
davidmrdavid, line 214: nit - same feedback regarding `default` and booleans

davidmrdavid (approved):
Approved with suggestions

src/DurableTask.Netherite/OrchestrationService/OrchestrationMessageBatch.cs
sebastianburckhardt: same answer. We are returning true, so the value of the out parameter is irrelevant.

PR #260, opened 2023-04-28
Fixes to azure storage calls

sebastianburckhardt:
Fixes a few omissions in code that uses the Azure Storage SDK:

1. do not report an error when retrying a delete and the blob was already deleted.
2. make sure to call Dispose() on all Azure responses that are IDisposable. This is a new requirement for Azure v12 SDK; previously, there was no such obligation on behalf of the caller.

Also, this PR removes a dependency on `FirstAsync` (a LINQ utility function for AsyncEnumberables) because this caused dependency issues for some .NET targets.

### DIFF
--- a/src/DurableTask.Netherite/Scaling/AzureTableLoadPublisher.cs
+++ b/src/DurableTask.Netherite/Scaling/AzureTableLoadPublisher.cs
@@ -38,7 +38,12 @@ namespace DurableTask.Netherite.Scaling
                 }
                 if (tableBatch.Count > 0)
                 {
-                    await this.table.SubmitTransactionAsync(tableBatch, cancellationToken).ConfigureAwait(false);
+                    var response = await this.table.SubmitTransactionAsync(tableBatch, cancellationToken).ConfigureAwait(false);
+
+                    foreach(var responseItem in response.Value)
+                    {
+                        responseItem.Dispose();
+                    }
                 }
             }
             catch(Azure.RequestFailedException e) when (e.Status == 404) // table may not exist
@@ -51,7 +56,7 @@ namespace DurableTask.Netherite.Scaling
             return this.table.CreateIfNotExistsAsync(cancellationToken);
         }
 
-        public Task PublishAsync(Dictionary<uint, PartitionLoadInfo> info, CancellationToken cancellationToken)
+        public async Task PublishAsync(Dictionary<uint, PartitionLoadInfo> info, CancellationToken cancellationToken)
         {
             var tableBatch = new List<TableTransactionAction>();
             foreach(var kvp in info)
@@ -60,11 +65,11 @@ namespace DurableTask.Netherite.Scaling
             }
             if (tableBatch.Count > 0)
             {
-                return this.table.SubmitTransactionAsync(tableBatch, cancellationToken);
-            }
-            else
-            {
-                return Task.CompletedTask;
+                var response = await this.table.SubmitTransactionAsync(tableBatch, cancellationToken);
+                foreach(var responseItem in response.Value)
+                {
+                    responseItem.Dispose();
+                }
             }
         }
 
--- a/src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
+++ b/src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
@@ -292,8 +292,16 @@ namespace DurableTask.Netherite.Faster
                     async (numAttempts) =>
                     {
                         var client = (numAttempts > 1) ? entry.PageBlob.Default : entry.PageBlob.Aggressive;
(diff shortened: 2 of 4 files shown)


src/DurableTask.Netherite/Scaling/AzureTableLoadPublisher.cs
```diff
@@ -38,7 +38,12 @@ await foreach (var e in this.table.QueryAsync<PartitionInfoEntity>(x => x.Partit
                 }
                 if (tableBatch.Count > 0)
                 {
-                    await this.table.SubmitTransactionAsync(tableBatch, cancellationToken).ConfigureAwait(false);
+                    var response = await this.table.SubmitTransactionAsync(tableBatch, cancellationToken).ConfigureAwait(false);
+
+                    foreach(var responseItem in response.Value)
+                    {
+                        responseItem.Dispose();
+                    }
```
davidmrdavid, line 46: similar to your changes elsewhere, could we replace this with a `using` statement?

src/DurableTask.Netherite/Scaling/AzureTableLoadPublisher.cs
```diff
@@ -60,11 +65,11 @@ public Task PublishAsync(Dictionary<uint, PartitionLoadInfo> info, CancellationT
...
-                return this.table.SubmitTransactionAsync(tableBatch, cancellationToken);
-            }
-            else
-            {
-                return Task.CompletedTask;
+                var response = await this.table.SubmitTransactionAsync(tableBatch, cancellationToken);
+                foreach(var responseItem in response.Value)
+                {
+                    responseItem.Dispose();
+                }
```
davidmrdavid, line 72: same feedback regarding `using`

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
```diff
@@ -292,8 +302,16 @@ public override void RemoveSegmentAsync(int segment, AsyncCallback callback, IAs
...
+                        try
+                        {
+                            using var response = await client.DeleteAsync(cancellationToken: this.PartitionErrorHandler.Token);
+                            return 1;
+                        }
+                        catch (Azure.RequestFailedException ex) when (numAttempts > 1 && BlobUtilsV12.BlobDoesNotExist(ex))
+                        {
+                            // blob may have already been deleted by the previous attempt
+                            return 0;
+                        }
```
davidmrdavid, line 314: Could we use a `TryDeleteIfExists`-type API instead?

davidmrdavid (commented):
Some questions

src/DurableTask.Netherite/Scaling/AzureTableLoadPublisher.cs
sebastianburckhardt: again, I see no technical way of using `using` as a means to call dispose here.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
sebastianburckhardt: Yes, that is a bit simpler. Will do.

davidmrdavid (approved):
LGTM

src/DurableTask.Netherite/Scaling/AzureTableLoadPublisher.cs
davidmrdavid: I don't have any great ideas here, unfortunately. My best suggestion would be to consider getting the IEnumerator or that IEnumerable (through GetEnumerator) and to simply call Dispose on that entire Enumerator (IEnumerator implements IDisposable) object in one go.

That would turn it into a 1 liner, but even then I don't think it's "better" than the explicit for loop. So yeah, no "great" ideas, just surprised there's not a more elegant alternative.

TedHartMS: I don't see a requirement that IEnumerator<T>.Dispose call Dispose() on its elements (see https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerator-1?view=net-8.0#notes-to-implementers); IEnumerator<T>.Dispose() is for things necessary to generate the sequence (connections, etc.). Which makes sense, especially if the elements are dynamically generated.

If you're doing this a lot you could write an extension for it.

sebastianburckhardt: > just surprised there's not a more elegant alternative.

Indeed. However, I don't think calling Dispose in a loop is the problem. The problem is that the whole idea of the programmer being responsible for calling `Dispose` on all `IDisposable` objects at the correct time is very tedious and error prone to begin with, and the `using` clause is just a bit of syntactic sugar to hide the bitter taste, and it will only work for the 'easy cases' where the object lifetime is statically scoped.

PR #261, opened 2023-04-28
limit frequency of failing partition starts

sebastianburckhardt:
I observed cases where partitions fail to start repeatedly. To reduce CPU waste and log pollution in those situations, this PR adds a backoff delay between retries.

### DIFF
--- a/src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsProcessor.cs
+++ b/src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsProcessor.cs
@@ -63,6 +63,7 @@ namespace DurableTask.Netherite.EventHubsTransport
             public TransportAbstraction.IPartition Partition;
             public Task<PartitionIncarnation> Next;
             public long NextPacketToReceive;
+            public int SuccessiveStartupFailures;
         }
 
         readonly Dictionary<string, MemoryStream> reassembly = new Dictionary<string, MemoryStream>();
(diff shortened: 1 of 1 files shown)


src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsProcessor.cs
```diff
@@ -155,7 +156,24 @@ async Task<PartitionIncarnation> StartPartitionAsync(PartitionIncarnation prior
                 {
                     // we are now becoming the current incarnation
                     this.currentIncarnation = prior.Next;
-                    this.traceHelper.LogDebug("EventHubsProcessor {eventHubName}/{eventHubPartition} is restarting partition (incarnation {incarnation}) soon", this.eventHubName, this.eventHubPartition, c.Incarnation);
+
+                    // sometimes we can get stuck into a loop of failing attempts to reincarnate a partition. 
+                    // We don't want to waste CPU  and pollute the logs, but we als can't just give up because
```
davidmrdavid, line 161:
```suggestion
                    // We don't want to waste CPU  and pollute the logs, but we also can't just give up because
```

src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsProcessor.cs
```diff
@@ -226,8 +244,9 @@ async Task<PartitionIncarnation> StartPartitionAsync(PartitionIncarnation prior
                 // the partition startup was canceled
                 this.traceHelper.LogDebug("EventHubsProcessor {eventHubName}/{eventHubPartition} canceled partition startup (incarnation {incarnation})", this.eventHubName, this.eventHubPartition, c.Incarnation);
             }
-            catch (Exception e) when (!Utils.IsFatal(e))
+            catch (Exception e)
```
davidmrdavid, line 247: Why did we remove the `IsFatal` check?

davidmrdavid (commented):
Some questions, almost ready to approve

src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsProcessor.cs
sebastianburckhardt: This snuck in when I was separating the PRs during my local development. It should actually be part of a different PR. Will remove from this one.

davidmrdavid (approved):

PR #262, opened 2023-04-28
fix LoadPublisher.FlushAsync

sebastianburckhardt:
The load publisher contained a bug where the `FlushAsync` basically only worked the first time, because it was using a single cancellationTokenSource. After that, the load publisher would run much too frequently. Also, this source was not properly disposed.

We fix this by using a TaskCompletionSource instead, and creating a new one for every invocation of FlushAsync.

### DIFF
--- a/src/DurableTask.Netherite/Scaling/LoadPublishWorker.cs
+++ b/src/DurableTask.Netherite/Scaling/LoadPublishWorker.cs
@@ -26,6 +26,8 @@ namespace DurableTask.Netherite.Scaling
 
         public Task FlushAsync()
         {
+            // wait for the worker to complete, but cancel the timed wait so it goes quicker
+            // correct order of these two is important, see discussion at https://github.com/microsoft/durabletask-netherite/pull/262#discussion_r1185209501// 
             var task = this.WaitForCompletionAsync();
             this.CancelCurrentWait();
             return task;
--- a/src/DurableTask.Netherite/Scaling/LoadPublishWorker.cs
+++ b/src/DurableTask.Netherite/Scaling/LoadPublishWorker.cs
@@ -64,13 +64,7 @@ namespace DurableTask.Netherite.Scaling
                 }
             }
 
-            try
-            {
-                await Task.WhenAny(Task.Delay(AggregatePublishInterval), this.cancelWait.Task).ConfigureAwait(false);
-            }
-            catch (OperationCanceledException)
-            {
-            }
+            await Task.WhenAny(Task.Delay(AggregatePublishInterval), this.cancelWait.Task).ConfigureAwait(false);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/DurableTask.Netherite/Scaling/LoadPublishWorker.cs
```diff
@@ -16,19 +16,26 @@ class LoadPublishWorker : BatchWorker<(uint, PartitionLoadInfo)>
...
             this.traceHelper = traceHelper;
-            this.cancelWait = new CancellationTokenSource();
         }
 
         public Task FlushAsync()
         {
-            this.cancelWait.Cancel(); // so that we don't have to wait the whole delay
-            return this.WaitForCompletionAsync();
+            var task = this.WaitForCompletionAsync();
+            this.CancelCurrentWait();
```
davidmrdavid, line 30: nit - can we flip the order of these two? It makes more sense to me that first we'd cancel the task, and then wait for completion

src/DurableTask.Netherite/Scaling/LoadPublishWorker.cs
```diff
@@ -59,7 +66,7 @@ protected override async Task Process(IList<(uint, PartitionLoadInfo)> batch)
 
             try
             {
-                await Task.Delay(AggregatePublishInterval, this.cancelWait.Token).ConfigureAwait(false);
+                await Task.WhenAny(Task.Delay(AggregatePublishInterval), this.cancelWait.Task).ConfigureAwait(false);
             }
             catch (OperationCanceledException)
             {
```
davidmrdavid, line 72: do we still need this catch statement? I think it's only for cancellation tokens.

davidmrdavid (commented):
Some questions

src/DurableTask.Netherite/Scaling/LoadPublishWorker.cs
sebastianburckhardt: Flipping the order would create a potential race: if calling it first, the `CancelCurrentWait` could cancel the current wait, the worker sees that there is no work, and starts waiting again (the worker is in an infinite loop that waits between iterations, and the cancel does not cancel the worker, just the wait). Now we call `WaitForCompletionAsync` and it has to wait for the whole duration because the cancel already canceled the previous wait.

In the current order, we are guaranteed that when we call `WaitForCompletionAsync`, the task has either already completed, or is queued for execution behind the wait.

(I think this was one of the bugs fixed in this PR).

src/DurableTask.Netherite/Scaling/LoadPublishWorker.cs
sebastianburckhardt: Good catch. I will remove the try/catch.

src/DurableTask.Netherite/Scaling/LoadPublishWorker.cs
davidmrdavid: Gotcha, that makes sense.
This seems subtle. Can you please add a comment here explaining the risk of this race? Otherwise, I worry we'll forget and may swap this in the future doing a refactoring.

davidmrdavid (approved):
Approved but please add a blub/comment explaining the race condition in FlushAsyc. This seems important to capture

PR #265, opened 2023-05-03
Add target based scaling support for Netherite

bachuv:
This PR adds support for target based scaling. It builds upon https://github.com/Azure/azure-functions-durable-extension/pull/2452. This PR is in draft mode since it hasn't been tested E2E yet.

Changes:
- [x] Add `NetheriteMetricsProvider` and move `CollectMetrics` code from ScaleMonitor here
- [x] Add `NetheriteTargetScaler` (implements `ITargetScaler`)
- [x] Override `TryGetTargetScaler` in NetheriteProvider
- [x] Add unit tests
- [ ] E2E testing


sebastianburckhardt:
I have spent some time to modify the logic so it now computes a target instead of a scale recommendation. It still has the same overall structure but I had to use the data provided by the metrics to determine the current scale.

bachuv:
Thanks Sebastian! I updated the PR and tested your changes locally.

bachuv:
@sebastianburckhardt, can you point me to where I should add unit tests?

sebastianburckhardt:
> @sebastianburckhardt, can you point me to where I should add unit tests?

I think the best place to add the tests is the DurableTask.Netherite.Tests project.

The other tests in there are mostly end-to-end tests that use some special fixture. In your case, just plain normal unit tests are probably enough, so I don't think you need to use any of those.

test/DurableTask.Netherite.AzureFunctions.Tests/TargetBasedScalingTests.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            this.metricsProviderMock = new Mock<NetheriteMetricsProvider>(
+                MockBehavior.Strict,
+                null,
+                new ConnectionInfo());
+        }
+
+        [Theory]
+        [InlineData(10, 10, 2)]
+        [InlineData(20, 20, 2)]
```
bachuv, line 41: @sebastianburckhardt, can you help provide expected input and output values to test the target based scaling algorithm? The inputs for this test are `maxConcurrentTaskActivityWorkItems` and `maxConcurrentTaskOrchestrationWorkItems`.

sebastianburckhardt:
Is this PR still needed?

davidmrdavid:
I believe we've parked this for the time being. Right @bachuv? I suspect this may become relevant again once Flex Consumption enter it's GA-prep phase, but we're not there yet.

@sebastianburckhardt: is it a problem if this remains open, just perhaps in draft mode?

bachuv:
> I believe we've parked this for the time being. Right @bachuv? I suspect this may become relevant again once Flex Consumption enter it's GA-prep phase, but we're not there yet.

@davidmrdavid, yes that's right.

davidmrdavid:
I've taken the liberty of marking this as draft to help signal that it doesn't need active review, and is simply an idle PR meant for future completion. Hope that's ok!

ericleigh007:
Not to be too obvious, but I guess this is why we were often peppered with log entries warning that target-based scaling could not be used.  The log entries are quite noisy.

davidmrdavid:
Hey @ericleigh007 - do you know if your app somehow "opted in" to target based scaling? If you're getting these warnings without opting in, that could be a problem (and possibly a reason to re-prioritize this work)

ericleigh007:
I was under the impression that target-based scaling was the default, and you had to opt-out.   So we have done nothing specifically for this.
This does seem to look different than the result we get with AzureStorage.
As another peripheral point, I thought that target-based scaling pivoted back to metric-based because mainly lack of management roles, but that may be an incorrect perception.

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
```diff
@@ -255,7 +255,7 @@ public bool TryGetScalingMonitor(out ScalingMonitor monitor)
         internal ILoadPublisherService GetLoadPublisher()
         {
             return string.IsNullOrEmpty(this.Settings.LoadInformationAzureTableName) ?
-                new AzureBlobLoadPublisher(this.Settings.BlobStorageConnection, this.Settings.HubName)
+                new AzureBlobLoadPublisher(this.Settings.BlobStorageConnection, this.Settings.HubName, this.Settings.TaskhubParametersFilePath)
```
bachuv, line 258: What is this additional parameter used for?

nytian: I am not sure haha, while the latest change at netherite just requires this

test/DurableTask.Netherite.AzureFunctions.Tests/DurableTask.Netherite.AzureFunctions.Tests.csproj
```diff
@@ -24,6 +24,7 @@
   <ItemGroup>
     <ProjectReference Include="..\..\src\DurableTask.Netherite.AzureFunctions\DurableTask.Netherite.AzureFunctions.csproj" />
     <ProjectReference Include="..\DurableTask.Netherite.Tests\DurableTask.Netherite.Tests.csproj" />
+	<PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="3.0.0" />
```
bachuv, line 27: nit: can we add this `PackageReference` in a separate/different `ItemGroup` since this `ItemGroup` only has `ProjectReference`s? Also, can we fix the indentation?

test/DurableTask.Netherite.Tests/DurableTask.Netherite.Tests.csproj
```diff
@@ -10,6 +10,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.11.1" />
+	<PackageReference Include="Moq" Version="4.18.4" />
```
sebastianburckhardt, line 13: nit: another indentation that is off. Try to use only spaces (no tabs) since tabs render differently in different environments.

test/DurableTask.Netherite.AzureFunctions.Tests/TargetBasedScalingTests.cs
sebastianburckhardt: I will take a look tomorrow. I think we may need more partitions also since we won't be able to test all decisions with only two partitions.

nytian (approved):

test/DurableTask.Netherite.AzureFunctions.Tests/TargetBasedScalingTests.cs
```diff
@@ -0,0 +1,100 @@
...
+        public async void TargetBasedScalingTest(int maxConcurrentTaskActivityWorkItems, int maxConcurrentTaskOrchestrationWorkItems, int expectedTargetWorkerCount)
+        {
+            this.orchestrationServiceMock.Setup(m => m.MaxConcurrentTaskActivityWorkItems).Returns(maxConcurrentTaskActivityWorkItems);
+            this.orchestrationServiceMock.Setup(m => m.MaxConcurrentTaskOrchestrationWorkItems).Returns(maxConcurrentTaskOrchestrationWorkItems);
+
+            var loadInformation = new Dictionary<uint, PartitionLoadInfo>()
+            {
+                { 1, this.Create("A") },
+                { 2, this.Create("B") },
+            };
```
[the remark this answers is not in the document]
sebastianburckhardt: Now replaced with a different method.

src/DurableTask.Netherite.AzureFunctions/NetheriteProvider.cs
```diff
@@ -130,6 +132,26 @@ public override bool TryGetScaleMonitor(
...
         }
 
+        public override bool TryGetTargetScaler(
+            string functionId,
+            string functionName,
+            string hubName,
+            string connectionName,
+            out ITargetScaler targetScaler)
+        {
+            ILoadPublisherService loadPublisher = this.Service.GetLoadPublisher();
```
sebastianburckhardt, line 142: This should be inside the conditional below, I think.

sebastianburckhardt (changes requested):
found one more thing to fix, otherwise looks good.

src/DurableTask.Netherite.AzureFunctions/NetheriteProvider.cs
```diff
@@ -130,6 +133,25 @@ public override bool TryGetScaleMonitor(
...
+            string functionId,
+            string functionName,
+            string hubName,
+            string connectionName,
+            out ITargetScaler targetScaler)
+        {           
+            // Target Scaler is created per function id. And they share the same NetheriteMetricsProvider.
+            if ( this.metricsProvider == null)
+            {
+                this.loadPublisher ??= this.Service.GetLoadPublisher();
```
nytian, line 146: Make loadPublisher class property too like metricsProvider so that it can be shared among scalers.

sebastianburckhardt (approved):

PR #283, opened 2023-07-06
Support specification of a consumer group

sebastianburckhardt:
This PR adds a feature to support specifying a consumer group other than the current default (`$Default`) in the options.

Specifying separate consumer groups makes it possible to run multiple task hubs on the same EH namespace without having them steal messages from each other.
Notes about this design:

- 

- to make this possible there is now a string option `EventHubsConsumerGroup` in `NetheriteOrchestrationServiceSettings` (and host.json). The default for this setting  is `$Default` which is also the default consumer group that event hubs already provides, and which is being used currently by all applications. Applications that do not explicitly change that setting will continue to function the same way as they do prior to this PR.

- If set to a value other than `$Default`, Netherite uses that consumer group instead of the default. It will also create this consumer group (which is an Azure resource) if it does not already exist. 

- Note that consumer grops are not automatically deleted, and some plans limit the number of consumer groups (lowest tier allows only a single consumer group). So if using the non-default consumer group, it is possible to get into a situation where Netherite cannot execute because it cannot create the consumer group. In that case users would have to explicitly delete some consumer groups (or perhaps just the entire EH namespace which is less complicated).
 
- Note that even if two task hubs that use the same EH namespace have different consumer groups, each message is still 'double delivered' since it is delivered to both task hubs (one of them will ignore it but it still consumes network and some CPU resources and counts towards the throughput limit). So, if both task hubs are heavily active at the same time, it may be a better idea to continue to use separate EH namespaces.

### DIFF
--- a/src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsConnections.cs
+++ b/src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsConnections.cs
@@ -274,7 +274,7 @@ namespace DurableTask.Netherite.EventHubsTransport
         }
 
 
-        public EventHubsSender<PartitionUpdateEvent> GetPartitionSender(int partitionId, byte[] taskHubGuid)
+        public EventHubsSender<PartitionUpdateEvent> GetPartitionSender(int partitionId, byte[] taskHubGuid, NetheriteOrchestrationServiceSettings settings)
         {
             return this._partitionSenders.GetOrAdd(partitionId, (key) => {
                 (EventHubClient client, string id) = this.partitionPartitions[partitionId];
@@ -284,13 +284,14 @@ namespace DurableTask.Netherite.EventHubsTransport
                     taskHubGuid,
                     partitionSender,
                     this.shutdownToken,
-                    this.TraceHelper);
+                    this.TraceHelper,
+                    settings);
                 this.TraceHelper.LogDebug("Created PartitionSender {sender} from {clientId}", partitionSender.ClientId, client.ClientId);
                 return sender;
             });
         }
 
-        public EventHubsClientSender GetClientSender(Guid clientId, byte[] taskHubGuid)
+        public EventHubsClientSender GetClientSender(Guid clientId, NetheriteOrchestrationServiceSettings settings)
         {
             return this._clientSenders.GetOrAdd(clientId, (key) =>
             {
@@ -304,11 +305,11 @@ namespace DurableTask.Netherite.EventHubsTransport
                 }
                 var sender = new EventHubsClientSender(
                         this.Host,
-                        taskHubGuid,
                         clientId,
                         partitionSenders,
                         this.shutdownToken,
-                        this.TraceHelper);
+                        this.TraceHelper,
+                        settings);
                 return sender;
             });
         }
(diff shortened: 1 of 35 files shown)


src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsConnections.cs
```diff
@@ -63,6 +68,11 @@ public async Task StartAsync(TaskhubParameters parameters)
                 this.EnsurePartitionsAsync(parameters.PartitionCount), 
                 this.EnsureClientsAsync(), 
                 this.EnsureLoadMonitorAsync());
+
+            if (this.consumerGroup != defaultConsumerGroup)
+            {
+                await this.EnsureConsumerGroupsExistAsync();
```
adarsh1, line 74: Should we consider doing the Consumer Group Creation in the OrchestrationServiceCreateAsync call? since its a resource being created?

well0549:
Well see discussion : https://github.com/microsoft/durabletask-netherite/discussions/292

I tried with the new setting : EventHubsConsumerGroup 

I added : AzureFunctionsJobHost__extensions__durableTask__EventHubsConsumerGroup  
to the configuration of the Function app, I did so for all 11 function apps, each with a different consumer name. After starting the functions the consumer groups were not created. Is there something i have to configure for that ?

sebastianburckhardt:
>  Is there something i have to configure for that ?

What type of authentication are you using? A SAS connection string or a managed identity?

> Furthermore, wouldn't is be more logical to have the taskhubname as the EventHubsConsumerGroup ?

Perhaps, yes, I can see that being a little simpler. 

We still need a configuration setting because this feature is not stable enough at this point to have it be always enabled.
Perhaps we could have a setting `EnableTaskHubConsumerGroups`.

well0549:
> > Is there something i have to configure for that ?
> 
> What type of authentication are you using? A SAS connection string or a managed identity?
> 
> > Furthermore, wouldn't is be more logical to have the taskhubname as the EventHubsConsumerGroup ?
> 
> Perhaps, yes, I can see that being a little simpler.
> 
> We still need a configuration setting because this feature is not stable enough at this point to have it be always enabled. Perhaps we could have a setting `EnableTaskHubConsumerGroups`.

I am using a managed identity for everythingf

sebastianburckhardt:
> I am using a managed identity for everything

Did you assign "Azure Event Hubs Data Owner" role? If not, it could be that the app does not have sufficient permissions to create the consumer group.

well0549:
> > I am using a managed identity for everything
> 
> Did you assign "Azure Event Hubs Data Owner" role? If not, it could be that the app does not have sufficient permissions to create the consumer group.

Yes I did

src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsConnections.cs
sebastianburckhardt: I think that would not work beause the event hub itself is dynamically created (and sometimes deleted and recreated for certain recovery scenarios).

sebastianburckhardt:
- change configuration setting to a boolean instead of a string with the consumer group name
- fix problem with partition event hub getting deleted every time a new task hub connects
- adjust the verbosity and contents of the logging

WendyCaiMS:
Hi @sebastianburckhardt, my team is looking to use this feature to support multiple task hubs. I see that this PR has been open for a while. Do you have any updates on this?

sebastianburckhardt:
We don't currently have a time estimate for providing this feature. 

The PR is unfortunately no longer up to date with the latest version of the event hubs SDK we are using now. Also, supporting multiple task hubs on the same EH namespace has some issues even when using consumer groups (e.g. the traffic is still redundantly delivered to all workers), and we were not able to fully automate the management of consumer groups for these scenarios, so we are a bit reluctant to add this feature since it will increase the support surface.

PR #290, opened 2023-07-21
Create unique scale monitor descriptors

bachuv:
This PR appends a GUID to the scale monitor descriptor to make it unique because customers have been seeing this exception - `System.ArgumentException : An item with the same key has already been added.`

This PR is used to create a private package for a customer to get this fix but it will get merged once the fix is validated.


sebastianburckhardt:
I thought this fix was already merged into the latest Netherite because I saw this PR which has status "merged": https://github.com/microsoft/durabletask-netherite/pull/270
But I realize now that this was only merged into another PR, not into the actual dev branch :(

Instead of doing a prerelease, why not an actual release? Since lots of customers are seeing this (I have also gotten lots of complaints).

davidmrdavid:
I see no issue with releasing this, although I don't love the GUID solution. Still, if we add a `#TODO:` comment to add the `functionId` to the string instead of a GUID (which should still be unique) then I would be good to merge. @bachuv, @sebastianburckhardt

PR #301, opened 2023-08-28
Terminate partition when FASTER refuses to checkpoint for over a minute

davidmrdavid:
Related: https://github.com/Azure/azure-functions-durable-extension/issues/2534

~This is a WIP, rough draft of adding logs to the checkpointing state machine so that we can better identify any bugs on this codepath.~

**Update:**

This PR contains a fix that terminates a partition when FASTER refuses to checkpoint for over a minute. In addition to this fix, we're adding more logs to the entire checkpoint state machine for easier debuggability moving forward. This PR is meant to be merged.


davidmrdavid:
This is not to be merged as-is, it is for a private package. If this logs prove to be useful there, we can look to format them in a way that is amenable for the prod-ready package.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -439,14 +451,17 @@ protected override async Task Process(IList<PartitionEvent> batch)
                 }
```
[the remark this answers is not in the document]
sebastianburckhardt: (want to rule out that the await does not resume even if the operation itself looks like it is done based on its own tracing)

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -320,6 +320,16 @@ bool CheckpointDue(out CheckpointTrigger trigger, out long? compactUntil)
...
+
+            var checkpointDueLog = $"CheckpointDue Debug: trigger={trigger}, " +
+                $"lastCheckpointedCommitLogPosition={this.lastCheckpointedCommitLogPosition}, " +
+                $"MaxNumberBytesBetweenCheckpoints={this.partition.Settings.MaxNumberBytesBetweenCheckpoints}," +
+                $"CommitLogPosition={this.CommitLogPosition}, " +
+                $"numberEventsSinceLastCheckpoint={this.numberEventsSinceLastCheckpoint}," +
+                $"MaxNumberEventsBetweenCheckpoints={this.partition.Settings.MaxNumberEventsBetweenCheckpoints}," +
+                $"inputQueuePositionLag={inputQueuePositionLag}," +
+                $"timeOfNextIdleCheckpoint={this.timeOfNextIdleCheckpoint}";
+            this.traceHelper.FasterProgress(checkpointDueLog);
```
sebastianburckhardt, line 332: This will generate A LOT of log messages so we won't want to merge this into the release but it is probably fine for a troubleshooting exercise.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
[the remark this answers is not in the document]
davidmrdavid: Good call. Added it.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
davidmrdavid: When you say "this" you mean this particular line, or the logs across this PR in general?

sebastianburckhardt: just this line. The others fire only as the checkpoint state machine progresses, which is not that frequent. But this line fires every time the storeworker goes through its loop body.

davidmrdavid: I opted to remove this. This generates ~5k logs per second per worker, which could cause ingestion problems or even overload the VM. I'm gonna look to make this log more periodic in a future iteration

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -4,6 +4,8 @@
 namespace DurableTask.Netherite.Faster
 {
     using DurableTask.Netherite.Scaling;
+    using Microsoft.Azure.Amqp.Framing;
+    using Microsoft.IdentityModel.Tokens;
```
davidmrdavid, line 8: not sure how this got added, will remove

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -349,7 +371,126 @@ void ScheduleNextIdleCheckpointTime()
...
+                if (durationFasterRefusal > TimeSpan.FromMinutes(1))
+                {
+                    messageOnError += $" FASTER first refused to checkpoint at '{this.startFasterCheckpointRefusal}'. Duration of refusal = {durationFasterRefusal}";
+                    Exception err = new Exception(messageOnError);
+                    this.partition.ErrorHandler.HandleError(nameof(StartCheckpointOrFailOnTimeout), messageOnError, err, terminatePartition: true, reportAsWarning: false);
+                }
+            }
+        }
+
+        async Task RunCheckpointingStateMachine()
```
sebastianburckhardt, line 399: You can make this a ValueTask to save an allocation on the common path (where the method completes synchronously)

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -349,7 +371,126 @@ void ScheduleNextIdleCheckpointTime()
...
+        {
+            // handle progression of checkpointing state machine:  none -> pendingCompaction -> pendingIndexCheckpoint ->  pendingStoreCheckpoint -> none)
+            if (this.pendingStoreCheckpoint != null)
```
davidmrdavid, line 402: I just moved this logic from `Process` onto this own method. But I realize that may make this PR harder to review. If it helps, I can undo this change and delay it to a separate PR.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -349,7 +371,126 @@ void ScheduleNextIdleCheckpointTime()
...
+                // if the refusal to checkpoint started over a minute ago, terminate partition
+                TimeSpan duration = currentTime - this.timeOfFirstRefusedCheckpoint.Value;
+                if (duration > TimeSpan.FromMinutes(1))
+                {
+                    messageOnError += $" FASTER first refused to checkpoint at '{this.timeOfFirstRefusedCheckpoint}'. Duration of refusal = {duration}";
+                    Exception err = new Exception(messageOnError);
+                    this.partition.ErrorHandler.HandleError(nameof(StartCheckpointOrFailOnTimeout), messageOnError, err, terminatePartition: true, reportAsWarning: false);
+                }
+            }
+        }
```
davidmrdavid, line 397: This is the fix

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -349,7 +371,126 @@ void ScheduleNextIdleCheckpointTime()
...
+                                this.traceHelper.FasterProgress("Checkpointing state machine: store checkpoint started");
+                                this.pendingIndexCheckpoint = null;
+                                this.pendingStoreCheckpoint = this.WaitForCheckpointAsync(false, token.Value, true);
+                                this.numberEventsSinceLastCheckpoint = 0;
+                            }
+                            var checkpointStarted = token.HasValue;
+                            return checkpointStarted;
+
+                        },
+                        messageOnError: "Could not start store checkpoint before timeout");
```
davidmrdavid, line 448: this is an example of the fix being applied in the state machine

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -349,7 +371,126 @@ void ScheduleNextIdleCheckpointTime()
...
+                // track start of FASTER refusal to start checkpoint
+                var currentTime = DateTime.UtcNow;
+                this.timeOfFirstRefusedCheckpoint ??= currentTime;
+
+                // if the refusal to checkpoint started over a minute ago, terminate partition
+                TimeSpan duration = currentTime - this.timeOfFirstRefusedCheckpoint.Value;
+                if (duration > TimeSpan.FromMinutes(1))
+                {
+                    messageOnError += $" FASTER first refused to checkpoint at '{this.timeOfFirstRefusedCheckpoint}'. Duration of refusal = {duration}";
+                    Exception err = new Exception(messageOnError);
```
davidmrdavid, line 393: I wasn't sure if we had more fine grained Exception classes for this kind of error, so I just used a generic `Exception` constructor

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -349,7 +371,126 @@ void ScheduleNextIdleCheckpointTime()
...
+                else
+                {
+                    this.traceHelper.FasterProgress("Checkpointing state machine: store checkpoint still pending");
+                }
+            }
+            else if (this.pendingIndexCheckpoint != null)
+            {
+                if (this.pendingIndexCheckpoint.IsCompleted == true)
+                {
+                    this.traceHelper.FasterProgress("Checkpointing state machine: trying to start store checkpoint");
```
sebastianburckhardt, line 429: suggested wording: pendingIndexCheckpoint has completed

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -349,7 +371,126 @@ void ScheduleNextIdleCheckpointTime()
...
+        }
+
+        async Task RunCheckpointingStateMachine()
+        {
+            // handle progression of checkpointing state machine:  none -> pendingCompaction -> pendingIndexCheckpoint ->  pendingStoreCheckpoint -> none)
+            if (this.pendingStoreCheckpoint != null)
+            {
+                if (this.pendingStoreCheckpoint.IsCompleted == true)
+                {
+                    this.traceHelper.FasterProgress("Checkpointing state machine: store checkpoint no longer pending");
```
sebastianburckhardt, line 406: suggested wording: pendingStoreCheckpoint has completed

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -349,7 +371,126 @@ void ScheduleNextIdleCheckpointTime()
...
+                            }
+                            var checkpointStarted = token.HasValue;
+                            return checkpointStarted;
+
+                        },
+                        messageOnError: "Could not start store checkpoint before timeout");
+                }
+                else
+                {
+                    this.traceHelper.FasterProgress("Checkpointing state machine: index checkpoint still pending");
```
sebastianburckhardt, line 452: Same as above; may make more sense to print this info as part of the `LogCheckpointStats`

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -349,7 +371,126 @@ void ScheduleNextIdleCheckpointTime()
...
+                    this.traceHelper.FasterProgress("Checkpointing state machine: resetting to initial state");
+                    // we have reached the end of the state machine transitions
+                    this.pendingStoreCheckpoint = null;
+                    this.pendingCheckpointTrigger = CheckpointTrigger.None;
+                    this.ScheduleNextIdleCheckpointTime();
+                    this.partition.Settings.TestHooks?.CheckpointInjector?.SequenceComplete((this.store as FasterKV).Log);
+                }
+                else
+                {
+                    this.traceHelper.FasterProgress("Checkpointing state machine: store checkpoint still pending");
```
sebastianburckhardt, line 422: This will again spam the logs. It may make more sense to print this info as part of the LogCheckpointStats

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -349,7 +371,126 @@ void ScheduleNextIdleCheckpointTime()
...
+                                this.pendingIndexCheckpoint = this.WaitForCheckpointAsync(true, token.Value, false);
+                            }
+                            var checkpointStarted = token.HasValue;
+                            return checkpointStarted;
+                        },
+                        messageOnError: "Could not start index checkpoint before timeout");
+                }
+                else
+                {
+                    this.traceHelper.FasterProgress("Checkpointing state machine: compaction still pending");
```
sebastianburckhardt, line 483: same as above.

only one of these pendingXXX can be non-null so you may save some space by just logging the name of the one that is hich one is non-null, if any

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -349,7 +371,126 @@ void ScheduleNextIdleCheckpointTime()
...
+                else
+                {
+                    this.traceHelper.FasterProgress("Checkpointing state machine: index checkpoint still pending");
+                }
+            }
+            else if (this.pendingCompaction != null)
+            {
+                if (this.pendingCompaction.IsCompleted == true)
+                {
+                    this.traceHelper.FasterProgress("Checkpointing state machine: trying to start index checkpoint");
```
sebastianburckhardt, line 459: suggested wording: pendingCompaction has completed

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
sebastianburckhardt: It's fine. It means i can't compare line by line but I think I can spot the differences since I looked at this code a lot in the last few days.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
sebastianburckhardt: You don't need to pass an exception, it is optional (only meant to be used if you caught one). Just pass null.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -349,7 +371,126 @@ void ScheduleNextIdleCheckpointTime()
...
+            {
+                // track start of FASTER refusal to start checkpoint
+                var currentTime = DateTime.UtcNow;
+                this.timeOfFirstRefusedCheckpoint ??= currentTime;
+
+                // if the refusal to checkpoint started over a minute ago, terminate partition
+                TimeSpan duration = currentTime - this.timeOfFirstRefusedCheckpoint.Value;
+                if (duration > TimeSpan.FromMinutes(1))
+                {
+                    messageOnError += $" FASTER first refused to checkpoint at '{this.timeOfFirstRefusedCheckpoint}'. Duration of refusal = {duration}";
```
sebastianburckhardt, line 392: nit: string is starting with an extra space
I dont' think you need to add quotes around a time.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
sebastianburckhardt: Got it. seems right.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -281,6 +282,28 @@ internal enum CheckpointTrigger
...
+
+            this.CheckpointDue(out CheckpointTrigger trigger, out long? compactUntil);
+            var checkpointDueLog = $"Checkpoint statistics: current checkpoint trigger={trigger}, " +
+                $"LastCheckpointedCommitLogPosition={this.lastCheckpointedCommitLogPosition}, " +
+                $"MaxNumberBytesBetweenCheckpoints={this.partition.Settings.MaxNumberBytesBetweenCheckpoints}," +
+                $"CommitLogPosition={this.CommitLogPosition}, " +
+                $"NumberEventsSinceLastCheckpoint={this.numberEventsSinceLastCheckpoint}," +
+                $"MaxNumberEventsBetweenCheckpoints={this.partition.Settings.MaxNumberEventsBetweenCheckpoints}," +
+                $"InputQueuePositionLag={inputQueuePositionLag}," +
+                $"TimeOfNextIdleCheckpoint={this.timeOfNextIdleCheckpoint}";
```
sebastianburckhardt, line 297: could add timeOfFirstRefusedCheckpoint here to help us get a lot of info in just one query.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
sebastianburckhardt: Since this is a rather large method, I would  consider making it an actual separate method instead of a local function. But only if it does not require you to pass a lot of context parameters.

sebastianburckhardt (commented):
First round of comments with minor suggestions.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
davidmrdavid: Good to know. I'll do that and also update the `HandleError` signature to flag that the Exception can be null.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
davidmrdavid: Yeah I recognize the phrasing here and in other similar logs was weird. I just wanted to avoid implying that every time this log was emitted, that the pendingStoreCheckpoint completed _again_, which isn't necessarily the case.

But now I think it's too minor of a concern. I'll take your suggestion.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
davidmrdavid: applied in next commit

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
davidmrdavid: applied in next commit

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
davidmrdavid: This is actually on purpose. Since we're adding new text to the error, I need to separate it with a space. I'm adding a dot (".") in there to make this clearer.

cgillum:
If this PR isn't meant to be merged, can we switch it to being a draft PR? We do this enough that I wonder if a new label might also be helpful.

davidmrdavid:
> If this PR isn't meant to be merged, can we switch it to being a draft PR? We do this enough that I wonder if a new label might also be helpful.

For visibility: we discussed this offline and clarified that this PR is meant to be merged. I also updated the description to make this clearer.

src/DurableTask.Netherite/OrchestrationService/PartitionErrorHandler.cs
```diff
@@ -61,7 +61,7 @@ public PartitionErrorHandler(int partitionId, ILogger logger, LogLevel logLevelL
             this.host = host;
         }
      
-        public void HandleError(string context, string message, Exception exception, bool terminatePartition, bool isWarning)
+        public void HandleError(string context, string message, Exception? exception, bool terminatePartition, bool isWarning)
```
cgillum, line 64: nit: Since we're making the `exception` parameter nullable here, I think we should do the same in the `TraceError` method.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -281,6 +282,42 @@ internal enum CheckpointTrigger
             Idle
         }
 
+        void LogCheckpointStats()
+        {
+            long inputQueuePositionLag = this.getInputQueuePositionLag();
+            this.CheckpointDue(out CheckpointTrigger trigger, out long? compactUntil);
+            var reportNullableTaskStatus = (Task? t) =>
```
cgillum, line 289: For improved performance, consider using a static local function instead of an `Action<Task>`. Something like:
```suggestion
            static string ReportNullableTaskStatus(Task? t)
```

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -281,6 +282,42 @@ internal enum CheckpointTrigger
...
+                $"CommitLogPosition={this.CommitLogPosition}, " +
+                $"NumberEventsSinceLastCheckpoint={this.numberEventsSinceLastCheckpoint}, " +
+                $"MaxNumberEventsBetweenCheckpoints={this.partition.Settings.MaxNumberEventsBetweenCheckpoints}, " +
+                $"InputQueuePositionLag={inputQueuePositionLag}," +
+                $"TimeOfNextIdleCheckpoint={this.timeOfNextIdleCheckpoint}, " +
+                $"TimeOfFirstRefusedCheckpoint={this.timeOfFirstRefusedCheckpoint}, " +
+                $"PendingCompaction status={reportNullableTaskStatus(this.pendingCompaction)}, " +
+                $"PendingIndexCheckpoint status={reportNullableTaskStatus(this.pendingIndexCheckpoint)}, " +
+                $"PendingStoreCheckpoint status={reportNullableTaskStatus(this.pendingStoreCheckpoint)}";
+            this.traceHelper.FasterProgress(checkpointDueLog);
```
cgillum, line 313: Are we expected to write this trace always or only when certain trace levels are configured? If the latter, and if we may end up writing this frequently, consider making this into a lambda function that gets passed to `this.traceHelper.FasterProgress` so that we execute this string generation only when needed (and put less pressure on the .NET GC).

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -281,6 +282,42 @@ internal enum CheckpointTrigger
...
+        }
+
+        long getInputQueuePositionLag()
```
cgillum, line 316: nit: method names in C# should always start with a capital letter.

cgillum (commented):
Haven't gone through everything yet, but here are a few initial comments.

src/DurableTask.Netherite/OrchestrationService/PartitionErrorHandler.cs
davidmrdavid: Indeed. Applied

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
davidmrdavid: Ah yes, I sometimes misremember and assume that only applies to public methods, but I guess it also applies to private ones. Applied this feedback.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
davidmrdavid: > Are we expected to write this trace always or only when certain trace levels are configured?

As per the definition of `FasterProgress`, this is only logged when the trace level is at least `Debug`. This is the case by default.

> If the latter, and if we may end up writing this frequently, consider making this into a lambda function that gets passed to this.traceHelper.FasterProgress

Good idea, I've incorporated that feedback in my next commit.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
davidmrdavid: Clever, didn't know we could create static _yet local_ methods. Incorporated in my next commit.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -281,6 +282,50 @@ internal enum CheckpointTrigger
...
+                else
+                {
+                    return t.IsCompleted ? "is completed" : "is not completed";
+                }
+            };
+
+            // since the statistics log is only emitted if the trace level is at least "Debug",
+            // we defer the construction of the string to relieve GC pressure
+            string ConstructLogString() {
+
```
cgillum, line 306: nit: in C#, the opening `{` is expected to be on its own line.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -561,6 +658,10 @@ public async Task<(long,(long,int))> WaitForCheckpointAsync(bool isIndexCheckpoi
 
                 this.partition.Settings.TestHooks?.CheckpointInjector?.CompactionComplete(this.partition.ErrorHandler);
             }
+            else
+            {
+                this.traceHelper.FasterProgress($"RunCompactionAsync: skipped due to 'null' target");
```
cgillum, line 663: Is there anything more we can say about the meaning of a "null" target? I don't expect a customer or a CSS engineer will understand what this means.

cgillum (approved):
Just a few minor nits but otherwise LGTM. Note that I didn't go too deep into the details as I'm not up-to-speed on how the core Netherite engine works, but the changes in this PR seem reasonable to me at a high level. I also assume that Sebastian has already provided sufficient feedback in terms of correctness (and I think I saw a note somewhere internally that this may have been validated by a customer?).

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
davidmrdavid: Yeah, I should have the muscle memory to remember this by now, thanks for the reminder. Applied on my next commit.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
davidmrdavid: I've refactored this method so that this log can be more informative. Please see the latest iteration

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -349,7 +394,120 @@ void ScheduleNextIdleCheckpointTime()
...
+                this.traceHelper.FasterProgress($"Checkpointing state machine: checkpoint is due. Trigger='{trigger}'. compactUntil='{compactUntil}'");
+                this.pendingCheckpointTrigger = trigger;
+
+                var isCompactionSkipped = !compactUntil.HasValue;
+                this.pendingCompaction = isCompactionSkipped ? Task.FromResult((long?)null) : this.RunCompactionAsync(compactUntil.Value);
+
+                if (isCompactionSkipped)
+                {
+                    this.traceHelper.FasterProgress($"RunCompactionAsync was skipped because a compaction range was not provided by trigger = {trigger}.");
+                }
```
davidmrdavid, line 507: @cgillum: the log informing the user that compaction was skipped was moved here btw. In this position, we're able to clarify that the reason why we skipped compaction is because a given compaction trigger did not specify a compaction range.

davidmrdavid:
@cgillum: thanks for your reviews. I'll still wait for a final LGTM from @sebastianburckhardt as well prior to merging this.

> and I think I saw a note somewhere internally that this may have been validated by a customer?

Right, a customer has been provided with the fix and they're currently monitoring whether or not it addressed the issue they reported. In any case, I'll wait for Sebastian's confirmation that this fix is sound before merging.

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
```diff
@@ -499,20 +503,39 @@ public override async Task<long> RunCompactionAsync(long target)
                     target - this.Log.BeginAddress,
                     this.GetElapsedCompactionMilliseconds());
 
+                var timeoutTask = Task.Delay(TimeSpan.FromMinutes(10));
```
davidmrdavid, line 506: Note: 10 minutes was determined to be a safe timeout window using various kusto queries to determine the expected duration of compaction. But it could be too conservative

sebastianburckhardt (changes requested):
Mostly pretty good but there are a few things that need fixing I think.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -281,6 +282,51 @@ internal enum CheckpointTrigger
             Idle
         }
 
+        void LogCheckpointStats()
+        {
+            long inputQueuePositionLag = this.GetInputQueuePositionLag();
+            this.CheckpointDue(out CheckpointTrigger trigger, out long? compactUntil);
```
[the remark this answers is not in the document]
davidmrdavid: Removed in my latest commit: https://github.com/microsoft/durabletask-netherite/pull/301/commits/df76d16022174c48503ab48c4fe34ef888136aa9

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
```diff
@@ -499,20 +503,39 @@ public override async Task<long> RunCompactionAsync(long target)
...
+                if (winner == timeoutTask)
+                {
+                    // compaction timed out. Terminate partition
+                    var exceptionMessage = $"Compaction {id} time out";
+                    this.partition.ErrorHandler.HandleError(nameof(RunCompactionAsync), exceptionMessage, e: null, terminatePartition: true, reportAsWarning: true);
+
+                    // we need resolve the task to ensure the 'finally' block is executed which frees up another thread to start compating
+                    tcs.SetException(new OperationCanceledException(exceptionMessage));
+                }
+
```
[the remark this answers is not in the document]
davidmrdavid: You're right, I cancel the timeout task in my latest commit just for good practice: https://github.com/microsoft/durabletask-netherite/pull/301/commits/df76d16022174c48503ab48c4fe34ef888136aa9

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
```diff
@@ -499,20 +503,39 @@ public override async Task<long> RunCompactionAsync(long target)
...
+                var winner = await Task.WhenAny(tcs.Task, timeoutTask);
+
+                if (winner == timeoutTask)
+                {
+                    // compaction timed out. Terminate partition
+                    var exceptionMessage = $"Compaction {id} time out";
+                    this.partition.ErrorHandler.HandleError(nameof(RunCompactionAsync), exceptionMessage, e: null, terminatePartition: true, reportAsWarning: true);
+
+                    // we need resolve the task to ensure the 'finally' block is executed which frees up another thread to start compating
+                    tcs.SetException(new OperationCanceledException(exceptionMessage));
```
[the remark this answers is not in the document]
davidmrdavid: Applied in latest commit: https://github.com/microsoft/durabletask-netherite/pull/301/commits/df76d16022174c48503ab48c4fe34ef888136aa9

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -553,15 +658,10 @@ public async Task<(long,(long,int))> WaitForCheckpointAsync(bool isIndexCheckpoi
...
-            if (target.HasValue)
-            {
-                target = await this.store.RunCompactionAsync(target.Value);
-
-                this.partition.Settings.TestHooks?.CheckpointInjector?.CompactionComplete(this.partition.ErrorHandler);
-            }
-
+            target = await this.store.RunCompactionAsync(target);
+            this.partition.Settings.TestHooks?.CheckpointInjector?.CompactionComplete(this.partition.ErrorHandler);
             this.Notify();
```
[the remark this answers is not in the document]
davidmrdavid: Address in earlier comment ,good catch!

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
```diff
@@ -349,7 +394,120 @@ void ScheduleNextIdleCheckpointTime()
...
+                        messageOnError: "Could not start index checkpoint before timeout");
+                }
+            }
+            else if (this.CheckpointDue(out var trigger, out long? compactUntil))
+            {
+                this.traceHelper.FasterProgress($"Checkpointing state machine: checkpoint is due. Trigger='{trigger}'. compactUntil='{compactUntil}'");
+                this.pendingCheckpointTrigger = trigger;
+
+                var isCompactionSkipped = !compactUntil.HasValue;
+                this.pendingCompaction = isCompactionSkipped ? Task.FromResult((long?)null) : this.RunCompactionAsync(compactUntil.Value);
```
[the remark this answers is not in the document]
davidmrdavid: In my latest commit, I move the `Notify` method out here so it is still called in both places. Does that seem like a fair refactoring?

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
```diff
@@ -503,7 +503,8 @@ public override async Task<long> RunCompactionAsync(long target)
                     target - this.Log.BeginAddress,
                     this.GetElapsedCompactionMilliseconds());
 
-                var timeoutTask = Task.Delay(TimeSpan.FromMinutes(10));
+                var tokenSource = new CancellationTokenSource();
```
sebastianburckhardt, line 506: `CancellationTokenSource` is an IDisposable and needs to be disposed at some point (otherwise you may be leaking resources such as kernel-provided synchronization primitives). See suggestion below on how to do that safely.

src/DurableTask.Netherite/StorageLayer/Faster/StoreWorker.cs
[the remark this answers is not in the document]
sebastianburckhardt: Not quite, unfortunately. When doing the compaction, the Notify needs to be called *after* the compaction completes. With your change, it gets called after the compaction is started, which is too soon.

(the point of Notify is to tell the storeworker that it is possible to move the state machine forward, which is true only if (a) we skipped compaction, or (b) compaction is complete)

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
```diff
@@ -517,7 +518,12 @@ public override async Task<long> RunCompactionAsync(long target)
...
 
                     // we need resolve the task to ensure the 'finally' block is executed which frees up another thread to start compating
-                    tcs.SetException(new OperationCanceledException(exceptionMessage));
+                    tcs.TrySetException(new OperationCanceledException(exceptionMessage));
+                }
+                else
+                {
+                    // cancel the timeout task since compaction completed
+                    tokenSource.Cancel();
                 }
```
sebastianburckhardt, line 527:
```suggestion
                }

                await timeoutTask.ContinueWith(_ => tokenSource.Dispose());
```

sebastianburckhardt (changes requested):

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
davidmrdavid: I wish VS would warn me about these. I'm sure it can, if I had the right linting rules. I'll take your suggestion, thanks

davidmrdavid:
@sebastianburckhardt: I see now why you wrote `RunCompactionAsync` in the way you did. I've reverted my changes to it in this commit: https://github.com/microsoft/durabletask-netherite/pull/301/commits/9158006f0757467835c210b42b936e99f96be646

sebastianburckhardt (approved):

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
```diff
@@ -504,7 +504,7 @@ public override async Task<long> RunCompactionAsync(long target)
                     this.GetElapsedCompactionMilliseconds());
 
                 var tokenSource = new CancellationTokenSource();
-                var timeoutTask = Task.Delay(TimeSpan.FromMinutes(10), tokenSource.Token); //.ContinueWith(_ => tokenSource.Dispose());
+                var timeoutTask = Task.Delay(TimeSpan.FromMinutes(10), tokenSource.Token).ContinueWith(_ => tokenSource.Dispose());
```
sebastianburckhardt, line 507: I don't think this is the right way to dispose the timer task.

The problem is that you want to dispose it after the timeout completes AND after line 526. Otherwise line 526 may throw an exception. 

That is why I suggested to put the ContinueWith on line 529.

sebastianburckhardt (changes requested):

sebastianburckhardt (approved):

PR #302, opened 2023-08-30
fix discard of packets so the next packet to expect is updated correctly

sebastianburckhardt:
Fixes a bug where discarding events from other taskhubs causes the expected sequence number to be wrong (not updated to reflect the discarded packets) which then triggers spurious errors when receiving more events.

This code is not tested yet.

### DIFF
--- a/src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsProcessor.cs
+++ b/src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsProcessor.cs
@@ -433,7 +433,11 @@ namespace DurableTask.Netherite.EventHubsTransport
                 {
                     this.traceHelper.LogDebug("EventHubsProcessor {eventHubName}/{eventHubPartition}({incarnation}) is processing packets", this.eventHubName, this.eventHubPartition, current.Incarnation);
 
-                    await foreach ((EventData eventData, PartitionEvent[] events, long seqNo) in this.blobBatchReceiver.ReceiveEventsAsync(this.taskHubGuid, packets, current.ErrorHandler.Token, current.NextPacketToReceive.seqNo))
+                    // we need to update the next expected seqno even if the iterator returns nothing, since it may have discarded some packets.
+                    // iterators do not support ref arguments, so we use a simple wrapper class to work around this limitation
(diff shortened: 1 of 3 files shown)


src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsProcessor.cs
```diff
@@ -433,7 +433,11 @@ async Task IEventProcessor.ProcessEventsAsync(PartitionContext context, IEnumera
                 {
                     this.traceHelper.LogDebug("EventHubsProcessor {eventHubName}/{eventHubPartition}({incarnation}) is processing packets", this.eventHubName, this.eventHubPartition, current.Incarnation);
 
-                    await foreach ((EventData eventData, PartitionEvent[] events, long seqNo) in this.blobBatchReceiver.ReceiveEventsAsync(this.taskHubGuid, packets, current.ErrorHandler.Token, current.NextPacketToReceive.seqNo))
+                    // we need to update the next expected seqno even if the iterator returns nothing, since it may have discarded some packets.
+                    // iterators do not support ref arguments, so we use a simple wrapper class to work around this limitation
+                    MutableLong nextPacketToReceive = new MutableLong() { Value = current.NextPacketToReceive.seqNo };
+
+                    await foreach ((EventData eventData, PartitionEvent[] events, long seqNo) in this.blobBatchReceiver.ReceiveEventsAsync(this.taskHubGuid, packets, current.ErrorHandler.Token, nextPacketToReceive))
```
davidmrdavid, line 440: We also call `ReceiveEventsAsync` in `LoadMonitorProcessor` and `EventHubsTransport`. What is the reason why we don't need to use update the expected `seqno` in there?

davidmrdavid (commented):
broadly makes sense, but have a quick question:

src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsProcessor.cs
sebastianburckhardt: Those two are used by clients and the load monitor, respectively. The latter do not care about the order or about duplication of received events, so they are not tracking the sequence numbers.

PR #305, opened 2023-09-15
Instrument event-log serialization/deserialization

sebastianburckhardt:
Several changes, to deal with recently observed fatal deserialization errors that can make partitions permanently unusable.

1. Adds detailed tracing (enabled only at lowest Trace level) that collects detailed information about each portion of data written to and read from the log.
2. Collects more information about the thrown deserialization exceptions.
3. Catches the deserialization exceptions and tries to continue as best as possible, ignoring the particular event that threw the exception.

### DIFF
--- a/src/DurableTask.Netherite/StorageLayer/Faster/LogWorker.cs
+++ b/src/DurableTask.Netherite/StorageLayer/Faster/LogWorker.cs
@@ -11,6 +11,7 @@ namespace DurableTask.Netherite.Faster
     using System.Threading.Tasks;
     using System.Threading.Channels;
     using FASTER.core;
+    using System.IO.Hashing;
 
     class LogWorker : BatchWorker<PartitionUpdateEvent>
     {
@@ -19,6 +20,7 @@ namespace DurableTask.Netherite.Faster
         readonly Partition partition;
         readonly StoreWorker storeWorker;
         readonly FasterTraceHelper traceHelper;
+        readonly bool traceLogDetails;
         bool isShuttingDown;
 
         readonly IntakeWorker intakeWorker;
@@ -34,6 +36,7 @@ namespace DurableTask.Netherite.Faster
             this.storeWorker = storeWorker;
             this.traceHelper = traceHelper;
             this.intakeWorker = new IntakeWorker(cancellationToken, this, partition.TraceHelper);
+            this.traceLogDetails = traceHelper.IsTracingAtMostDetailedLevel;
 
             this.maxFragmentSize = (int) this.blobManager.GetEventLogSettings(partition.Settings.UseSeparatePageBlobStorage, partition.Settings.FasterTuningParameters).PageSize - 64; // faster needs some room for header, 64 bytes is conservative
         }
@@ -112,7 +115,7 @@ namespace DurableTask.Netherite.Faster
         {
             if (bytes.Length <= this.maxFragmentSize)
             {
-                this.log.Enqueue(bytes);
+                this.AppendToLog(bytes);
             }
             else
             {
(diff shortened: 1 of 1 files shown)


src/DurableTask.Netherite/StorageLayer/Faster/LogWorker.cs
```diff
@@ -123,12 +126,30 @@ void AddToFasterLog(byte[] bytes)
...
         }
 
+        void AppendToLog(ReadOnlySpan<byte> span)
+        {
+            this.log.Enqueue(span);
+            if (this.traceLogDetails)
+            {
+                this.TraceLogDetail("Appended", this.log.TailAddress, span);
+            }
+        }
```
davidmrdavid, line 142: wait so these are the literal bytes appended to the FASTER log? Could this contain private data?

src/DurableTask.Netherite/StorageLayer/Faster/LogWorker.cs
```diff
@@ -123,12 +126,30 @@ void AddToFasterLog(byte[] bytes)
...
+
+        void TraceLogDetail(string operation, long nextCommitLogPosition, ReadOnlySpan<byte> span)
+        {
+            byte firstByte = span[0];
+            char f = ((firstByte & first) != 0) ? 'F' : '.';
+            char l = ((firstByte & last) != 0) ? 'L' : '.';
+            byte[] crc = Crc32.Hash(span);
+            this.traceHelper.FasterStorageProgress($"CommitLogEntry {operation} {nextCommitLogPosition} {f}{l} length={span.Length} crc={crc[0]:X2}{crc[1]:X2}{crc[2]:X2}{crc[3]:X2}");
```
davidmrdavid, line 150: I assume this CRC is related to error correction codes. Why do we need to log this and how can it help us debug what went wrong?

src/DurableTask.Netherite/StorageLayer/Faster/LogWorker.cs
```diff
@@ -264,11 +285,17 @@ async IAsyncEnumerable<PartitionUpdateEvent> EventsToReplay(long from)
...
+                    {
+                        this.TraceLogDetail("Read", iter.NextAddress, new ReadOnlySpan<byte>(result, 0, entryLength));
+                    }
+
                     if ((result[0] & first) != none)
                     {
                         if ((result[0] & last) != none)
                         {
-                            partitionEvent = (PartitionUpdateEvent)Serializer.DeserializeEvent(new ArraySegment<byte>(result, 1, entryLength - 1));
+                            partitionEvent = TryDeserializePartitionEvent(new MemoryStream(result, 1, entryLength - 1), reassembled:false);
```
davidmrdavid, line 297: I haven't fully grok'ed this codepath, but it appears to me that a single partition event can sometimes be split into several pieces, which are then re-eassmebled using the memoryStream APIs. Is that what this `reassembled` parameter is representing? Essentially, at this point in the code, we haven't fully re-assembled an atomic partition event?

src/DurableTask.Netherite/StorageLayer/Faster/LogWorker.cs
davidmrdavid: In particular, also use some clarity this `first` and `last` comparisons. Not terribly familiar with this byte checks.

davidmrdavid (commented):
A few questions

src/DurableTask.Netherite/StorageLayer/Faster/LogWorker.cs
sebastianburckhardt: The `TraceLogDetail` method is not tracing the actual contents, just information about it.

src/DurableTask.Netherite/StorageLayer/Faster/LogWorker.cs
sebastianburckhardt: I am just using the CRC as an easy way to do checksum. Means I can check whether FASTER is reading back the same bytes as it wrote.

first and last are used by the fragmentation/reassembly logic. When writing a packet that is too large for FASTER it gets split and the first packet has a FIRST flag, the last packed has a LAST flag. If there is only one packet it has both a FIRST and LAST flag.

src/DurableTask.Netherite/StorageLayer/Faster/LogWorker.cs
sebastianburckhardt: yes, see notice above about fragmentation and reassembly. We reassemble all the pieces when we receive the last portion.

src/DurableTask.Netherite/StorageLayer/Faster/LogWorker.cs
davidmrdavid: Thanks. As a nit: can you please write this information down as a comment in this method? I think it's useful context

src/DurableTask.Netherite/StorageLayer/Faster/LogWorker.cs
davidmrdavid: Got it. So - since we're trying to reconstruct an event from various pieces, if any of them fails to deserialize and there's still more pieces to assemble, shouldn't we abort trying to deserialize the next piece/fragment of this event?

sebastianburckhardt: We are only trying to deserialize once we got the last portion.

davidmrdavid: Ah, I see that now. Goes on to show that I did not understand this codepath, but I see how it works now. Sounds good, no blocker here then.

davidmrdavid (approved):
LGTM so long as this is included: https://github.com/microsoft/durabletask-netherite/pull/305/files#r1329328310

davidmrdavid:
FYI releasing private package from this forked branch: https://github.com/microsoft/durabletask-netherite/compare/pr/fix-and-instrument-commitlog-deserialization...dajusto/release-deserialization-as-private-package?expand=1

PR #309, opened 2023-09-27
Copy ParentTraceContext when cloning ExecutionStartedEvent

bachuv:
This PR adds a line to copy `ExecutionStartedEvent.ParentTraceContext` when cloning `ExecutionStartedEvent` in `CreationRequestReceived.Clone()`. This change is necessary for Distributed Tracing V2 to work.

I created this PR as a draft because a new version of DT.Core needs to be released with a bug fix for this to work so this PR shouldn't be merged until then.


sebastianburckhardt (approved):
LGTM, thanks!

src/DurableTask.Netherite/Events/PartitionEvents/External/FromClients/CreationRequestReceived.cs
```diff
@@ -75,6 +75,7 @@ public override PartitionEvent Clone()
                     ExtensionData = ee.ExtensionData,
                     OrchestrationInstance = ee.OrchestrationInstance,
                     Timestamp = this.CreationTimestamp,
+                    ParentTraceContext = ee.ParentTraceContext,
                 },
                 SequenceNumber = tm.SequenceNumber,
                 OrchestrationInstance = tm.OrchestrationInstance,
```
davidmrdavid, line 81: non blocking question: it appears this copy of the execution started event is very manual, so I worry that we'll break it again in the future if we add a new field. Is there a more future-proof way of doing this copy? For example, via [Object.MemberwiseClone()](https://learn.microsoft.com/en-us/dotnet/api/system.object.memberwiseclone?view=net-7.0)?

If so, perhaps we can add a TODO comment on this line to object copy to consider changing the cloning approach

bachuv: @sebastianburckhardt, what are your thoughts on this?

bachuv:
@sebastianburckhardt, can we merge this PR to be a part of the v1.4.1 release?

src/DurableTask.Netherite/Events/PartitionEvents/External/FromClients/CreationRequestReceived.cs
sebastianburckhardt: Using `Object.MemberwiseClone` seems like a good idea to me.

bachuv:
Leaving a note that this PR depends on DT.Core > 2.15.1.

sebastianburckhardt (approved):

Issue #310, opened 2023-09-28
Replace Microsoft.Azure.EventHubs.Processor with Azure.Messaging.EventHubs.Processor


cgillum:
This is required for internal Microsoft component governance ([internal Microsoft link](https://msazure.visualstudio.com/One/_componentGovernance/AAPT-Antares-ScaleController/alert/8964806?typeId=16985491)).

The problem is that `Microsoft.Azure.EventHubs.Processor` (which has been deprecated) has a dependency on old ADAL libraries, and there is a high-level mandate to get rid of all ADAL usage across the company for security and compliance reasons.

The compliant version of the event hub processor can be found here: https://www.nuget.org/packages/Azure.Messaging.EventHubs.Processor/

Importantly, the scale controller will also need to be updated with a new build that uses the updated event hubs processor. Alternatively, if the scale controller doesn't require the Event Hubs processor (I assume it doesn't), we can create a version of the Netherite client library that doesn't include this dependency.

sebastianburckhardt:
FYI, some Netherite users have also asked about this since this deprecation is causing them similar issues.

We will need to provide some guidance/timeline as to when we expect to be able to remove this dependency.

NickCraver:
Adding info here: I tried to do a spike migration to see how much effort would be involved here. Here's a quick link to the guide as a ref as well: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/MigrationGuide.md

Overall:
- The way clients work has fundamentally changed - they aren't per-partition anymore but are send/receive with control of partition within as well as some for processing and such (our main case, I think). This affects everything from connection strings down.
  - Subset: Partition shutdowns are different so we'd have to handle shutdown differently too - this may get into lifetime mechanics and become very fun
- Event processors are gone in Azure.Messaging.EventHubs (things are not event based as hookups instead of implementing `IEventProcessor`
- Offsets/SequenceNumber/etc. have moved but are still there. In general though I think the iteration mechanics are meant to be IEnumerable/handled for you in most cases where a lot of it's manual in code today. This should be net cleaner, but isn't a quick port
- `PartitionEvent`s are not structs, not subclassed
- General close mechanics differ overall (unsure of the extent, but reasons/handling differ for sure)
- Besides those, lots of misc metadata changes.

[...]

PR #311, opened 2023-09-29
[Projects] Simplify shared props & split Microsoft.Azure.EventHubs.Processor dependency

NickCraver:
#### Draft PR for discussion of #310

I figured this was the easiest way to envision what a split looks like, so starting a PR to help discussion on what ties remain problematic.

------

This is the hopefully happy path proposal of splitting the dependency to defer the upgrade. 

Overall changes:
- Utilized Directory.Build.props to DRY up projects (and misc. associated changes like one icon.png file, etc.)
- Creates a new DurableTask.Netherite.EventHubs project which is the only thing depending on Microsoft.Azure.EventHubs.Processor.
- Moves all EventHubs code possible to DurableTask.Netherite.EventHubs

The problem here is there are still a few refs in core classes we have to figure out:
- `ScalingMonitor` is calling `EventHubsConnections.GetQueuePositionsAsync` to see if work is left
- `FasterStorageProvider` is calling `EventHubsUtil.DeleteEventHubIfExistsAsync` to cleanup partition hubs to ensure queues start at zero

image

Other than those 2 usages, the split is pretty clean here. I do recommend we fully drop `netcoreapp2.2` though - I assume based on comments and other cleanup that has happened it's explicitly here for a reason I don't understand, happy to understand that and maybe advise on some alternatives to clean up the build if we still need to keep it around.


NickCraver:
cc @cgillum @sebastianburckhardt

sebastianburckhardt:
I think we now have addressed this. The most urgent problem (scale controller conformance) has been solved by updating the scale controller selectively, I believe. As for the rest of the codebase, we now have a PR that updates all uses of the SDK here: https://github.com/microsoft/durabletask-netherite/pull/385

PR #313, opened 2023-10-11
Rev version for 1.4.1 release and increase DF + DTFx deps

davidmrdavid:
Increasing DF Extension to 2.11.4, which also increases the WebJobs dependency. Also increasing the DTFx.Core version accordingly


davidmrdavid:
No longer needed.

Issue #325, opened 2023-10-23
Exceptions due to concurrent modification of HistoryState


sebastianburckhardt:
Observing occasional failures in CI with this type of stack trace:

```
[xUnit.net 00:15:28.13]         2023-10-18T22:04:31.7597959+00:00 Part00 !!! Encountered exception while working on store in StoreWorker.Process: System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
[xUnit.net 00:15:28.13]            at WriteArrayOfHistoryEventToXml(XmlWriterDelegator , Object , XmlObjectSerializerWriteContext , CollectionDataContract )
[xUnit.net 00:15:28.13]            at System.Runtime.Serialization.CollectionDataContract.WriteXmlValue(XmlWriterDelegator xmlWriter, Object obj, XmlObjectSerializerWriteContext context)
[xUnit.net 00:15:28.13]            at System.Runtime.Serialization.XmlObjectSerializerWriteContext.WriteDataContractValue(DataContract dataContract, XmlWriterDelegator xmlWriter, Object obj, RuntimeTypeHandle declaredTypeHandle)
[xUnit.net 00:15:28.13]            at System.Runtime.Serialization.XmlObjectSerializerWriteContext.SerializeWithoutXsiType(DataContract dataContract, XmlWriterDelegator xmlWriter, Object obj, RuntimeTypeHandle declaredTypeHandle)
[xUnit.net 00:15:28.13]            at System.Runtime.Serialization.XmlObjectSerializerWriteContext.InternalSerialize(XmlWriterDelegator xmlWriter, Object obj, Boolean isDeclaredType, Boolean writeXsiType, Int32 declaredTypeID, RuntimeTypeHandle declaredTypeHandle)
[xUnit.net 00:15:28.13]            at WriteHistoryStateToXml(XmlWriterDelegator , Object , XmlObjectSerializerWriteContext , ClassDataContract )
[xUnit.net 00:15:28.13]            at System.Runtime.Serialization.ClassDataContract.WriteXmlValue(XmlWriterDelegator xmlWriter, Object obj, XmlObjectSerializerWriteContext context)
[xUnit.net 00:15:28.13]            at System.Runtime.Serialization.XmlObjectSerializerWriteContext.WriteDataContractValue(DataContract dataContract, XmlWriterDelegator xmlWriter, Object obj, RuntimeTypeHandle declaredTypeHandle)
[...]

PR #326, opened 2023-10-23
Fix fencing of checkpoints

sebastianburckhardt:
This PR fixes a problem with how Netherite uses the asynchronous checkpointing support in FASTER.

The problem was that there were races between the checkpoint thread and the main session.
@badrishc  suggested to fix this by waiting, after initiating the checkpoint, for the main session to advance to the next version.

I added that loop, together with some safety limit on how long to wait.

Fixes https://github.com/microsoft/durabletask-netherite/issues/325

### DIFF
--- a/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
+++ b/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
@@ -218,7 +218,14 @@ namespace DurableTask.Netherite.Faster
         {
             this.singletons = new TrackedObject[TrackedObjectKey.NumberSingletonTypes];
             string suffix = DateTime.UtcNow.ToString("O");
+            // The main session is the session used by all reads and writes performed by the storeworker.
+            // Using a single session means that the store worker sees a consistent state at all times.
+            // A single session is sufficient since the storeworker performs only one operation at a time.
             this.mainSession = this.CreateASession($"main-{suffix}", false);
+            // Since queries may require scans that can take a significant time, it is not advisable to perform
+            // queries on the main session. We therefore create a pool of read-only sessions available for queries. 
+            // Query sessions may see a slightly stale store than the main session, i.e. may not contain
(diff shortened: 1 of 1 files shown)


src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
```diff
@@ -416,10 +416,28 @@ public override async Task FinalizeCheckpointCompletedAsync(Guid guid)
                     this.fht.Log.ShiftBeginAddress(shiftBeginAddress.Value);
                 }
 
+                long versionBeforeCheckpoint = this.mainSession.Version;
+
```
davidmrdavid, line 420: Just for my understanding: here the `session` refers to what in FASTER? Is it an ongoing process, the state before checkpointing in storage, or something else? I'd appreciate having that as an inline comment in the code for future reference.

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
```diff
@@ -416,10 +416,28 @@ public override async Task FinalizeCheckpointCompletedAsync(Guid guid)
...
                 if (this.fht.TryInitiateHybridLogCheckpoint(out var token, CheckpointType.FoldOver))
                 {
                     // according to Badrish this ensures proper fencing w.r.t. session
-                    this.mainSession.Refresh();
+                    Stopwatch stopwatch = Stopwatch.StartNew();
+                    TimeSpan timeLimit = TimeSpan.FromMinutes(1);
+                    while (stopwatch.Elapsed < timeLimit)
+                    {
+                        this.mainSession.Refresh();
```
davidmrdavid, line 428: does this mean: fetch the latest session info? Let's comment inline if so

davidmrdavid (commented):
Some nits for documentation's sake

davidmrdavid (approved):

Issue #312, opened 2023-10-07
Repeating Timers on no longer existing durable function and instance


rmbeaty:
Hi,

I've am using Netherite in Azure Functions. At one point I had a durable functions that repeated on a durable timer. However that class was deleted from the source, but now I am still seeing some type of activity for it in the Log Info, subset of these below (these seem to repeat every few seconds)....

2023-10-07T13:39:28Z   [Information]   Part05.21148384599712 Replayed internal UpdateEvent TimerFired eventId=05T147841722 instanceId=@durableaiimagegenerator@AIImageGenerator_0 pos=(21148384599712,0) latency=(0, 49094, 0)
2023-10-07T13:39:28Z   [Information]   Part05.21148384601104 Replayed internal UpdateEvent TimerFired eventId=05T147841723 instanceId=@durableaiimagegenerator@AIImageGenerator_0 pos=(21148384601104,0) latency=(0, 49094, 0)
2023-10-07T13:39:28Z   [Information]   Part05.21148384602496 Replayed internal UpdateEvent TimerFired eventId=05T147841724 instanceId=@durableaiimagegenerator@AIImageGenerator_0 pos=(21148384602496,0) latency=(0, 49094, 0)
2023-10-07T13:39:28Z   [Information]   Part05.21148384603884 Replayed internal UpdateEvent TimerFired eventId=05T147841725 instanceId=@durableaiimagegenerator@AIImageGenerator_0 pos=(21148384603884,0) latency=(0, 49094, 0)
2023-10-07T13:39:28Z   [Information]   Part05.21148384605276 Replayed internal UpdateEvent TimerFired eventId=05T147841726 instanceId=@durableaiimagegenerator@AIImageGenerator_0 pos=(21148384605276,0) latency=(0, 49094, 0)
2023-10-07T13:39:28Z   [Information]   Part05.21148384606668 Replayed internal UpdateEvent TimerFired eventId=05T147841727 instanceId=@durableaiimagegenerator@AIImageGenerator_0 pos=(21148384606668,0) latency=(0, 49094, 0)
2023-10-07T13:39:28Z   [Information]   Part05.21148384608060 Replayed internal UpdateEvent TimerFired eventId=05T147841728 instanceId=@durableaiimagegenerator@AIImageGenerator_0 pos=(21148384608060,0) latency=(0, 49094, 0)
[...]

PR #329, opened 2023-10-31
Support automatic resizing for variable-size page blobs

sebastianburckhardt:
Our current implementation was allocating a default of 512GB for variable-sized page blobs. This was considered sufficient at the time since these blobs are by design not growing forever (they are used for checkpoints and for object logs, both of which have a limited life time).

However, it turns out that it is not always enough (https://github.com/microsoft/durabletask-netherite/issues/312).

In this PR, I revised the mechanism so that 

1. when a variable-sized page blob runs out of space, I am enlarging it to double its size.
2. when a read attempts to read past the end of the blob, I cap the read so it stays within the size bounds of the page blob.

I have tested this manually by reducing the starting size to something very small, and running a sample workload on it.

### DIFF
--- a/src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
+++ b/src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
@@ -62,7 +62,10 @@ namespace DurableTask.Netherite.Faster
         const uint MAX_UPLOAD_SIZE = 1024 * 1024;
         const uint MAX_DOWNLOAD_SIZE = 1024 * 1024;
 
-        const long MAX_PAGEBLOB_SIZE = 512L * 1024 * 1024 * 1024; // set this at 512 GB for now TODO consider implications
+        // some blobs have no specific size; for example, they are used for the object log, and for checkpoints.
+        // we start them at 512 GB - there is no cost incurred for empty pages so why not.
+        // if that turns out to not be enough at some point, we enlarge the page blob automatically.
+        const long STARTING_DEFAULT_PAGEBLOB_SIZE = 512L * 1024 * 1024 * 1024; 
 
         /// <summary>
         /// Constructs a new AzureStorageDevice instance, backed by Azure Page Blobs
@@ -425,7 +428,7 @@ namespace DurableTask.Netherite.Faster
                     var pageBlob = this.pageBlobDirectory.GetPageBlobClient(this.GetSegmentBlobName(segmentId));
 
                     // If segment size is -1 we use a default
-                    var size = this.segmentSize == -1 ? AzureStorageDevice.MAX_PAGEBLOB_SIZE : this.segmentSize;
+                    var size = this.segmentSize == -1 ? AzureStorageDevice.STARTING_DEFAULT_PAGEBLOB_SIZE : this.segmentSize;
 
                     // If no blob exists for the segment, we must first create the segment asynchronouly. (Create call takes ~70 ms by measurement)
                     // After creation is done, we can call write.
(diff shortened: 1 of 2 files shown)


src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
```diff
@@ -516,31 +556,83 @@ async Task ReadFromBlobAsync(UnmanagedMemoryStream stream, BlobUtilsV12.PageBlob
...
+                                    }
+                                    else if (actualLength > requestedLength)
+                                    {
+                                        throw new InvalidDataException($"PageBlob.DownloadStreamingAsync returned too much data, range={response.Value.Details.ContentRange} requestedLength={requestedLength} actualLength=${actualLength}");
+                                    }
+                                }
+                                catch (Azure.RequestFailedException e) when (e.ErrorCode == "InvalidRange")
+                                {
+                                    // this type of error can be caused by the page blob being smaller than where FASTER is reading from.
+                                    // to handle this, first determine current page blob size.
```
davidmrdavid, line 620: not sure I fully understand this comment. What does it mean to be smaller than _where_ FASTER is reading from? Can we add more detail?

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
```diff
@@ -500,11 +534,17 @@ async Task ReadFromBlobAsync(UnmanagedMemoryStream stream, BlobUtilsV12.PageBlob
         {
             using (stream)
             {
+                // we use this to prevent reading past the end of the page blob
```
davidmrdavid, line 537: I don't immediately understand why enlarging the blobs can suddenly cause reads to go past the end of the page blob. Can you please call out the connection?

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
```diff
@@ -516,31 +556,83 @@ async Task ReadFromBlobAsync(UnmanagedMemoryStream stream, BlobUtilsV12.PageBlob
...
-                                    await streamingResult.Content.CopyToAsync(stream).ConfigureAwait(false);
+                                    requestedLength = 0;
                                 }
                             }
 
-                            if (stream.Position != offset + length)
+                            if (requestedLength > 0)
                             {
-                                throw new InvalidDataException($"wrong amount of data received from page blob, expected={length}, actual={stream.Position}");
+                                var client = (numAttempts > 1 || requestedLength == MAX_DOWNLOAD_SIZE) ? blob.Default : blob.Aggressive;
```
davidmrdavid, line 575: can you comment on what this controls? What is `blob.Aggressive`?

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
```diff
@@ -516,31 +556,83 @@ async Task ReadFromBlobAsync(UnmanagedMemoryStream stream, BlobUtilsV12.PageBlob
...
+                                        cancellationToken: this.PartitionErrorHandler.Token)
+                                        .ConfigureAwait(false);
+
+                                    using (var streamingResult = response.Value)
+                                    {
+                                        await streamingResult.Content.CopyToAsync(stream).ConfigureAwait(false);
+                                    }
+
+                                    // we have observed that we may get 206 responses where the actual length is less than the requested length
+                                    // this seems to only happen if blob was enlarged in the past
```
davidmrdavid, line 592: So what are we doing if we get a 206? I actually don't see us checking the response status explicitly, should we?
nit: let's also add in parenthesis the meaning of 206: "partial response", right?

davidmrdavid (commented):
Thanks for the quick solution here. I have various nits and questions as I think this is a non-trivial fix. I think adding a few more comments here will help with readability and future maintenance

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
```diff
@@ -469,15 +472,46 @@ async Task WritePortionToBlobAsync(UnmanagedMemoryStream stream, BlobEntry blobE
...
+                            {
+                                // this kind of error can indicate that the page blob is too small.
+                                // first, compute desired size to request
+                                long currentSize = (await client.GetPropertiesAsync().ConfigureAwait(false)).Value.ContentLength;
+                                long sizeToRequest = currentSize; 
+                                long sizeToAccommodate = destinationAddress + offset + length + 1;
+                                while (sizeToAccommodate > sizeToRequest)
+                                {
+                                    sizeToRequest <<= 1;
+                                }
```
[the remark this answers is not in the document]
sebastianburckhardt: The page size increase does already get logged as a warning, because the ForceRetryException that we throw.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
sebastianburckhardt: Because these blobs don't actually have a fixed size from the perspective of FASTER, the FASTER reads may easily read past the end of the blob. This is not new - we just never saw it because 512GB was always way larger than needed.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
sebastianburckhardt: they control the length of the timeout (with newer storage SDK the timeout can no longer be passed as a parameter to the operation, since it is a constructor parameter for the client, so we need multiple clients to control the timeout length).

BTW this was not modified in this PR, the diff just does not show this well.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
sebastianburckhardt: The SDK internally takes care of interpreting the status response when I access the streaming response (`response.Value`). From my perspective, what I see in this case is that the stream returned has fewer bytes than requested.

BTW, I have seen a few 206 during testing early on but later no longer saw them. I figured it makes sense to handle them just in case they pop up. It is pretty easy to handle them by just adjusting the next portion to start where the 206 ended.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
sebastianburckhardt: FASTER may be reading a range that extends past the end of the blob.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
```diff
@@ -425,7 +428,7 @@ public override void WriteAsync(IntPtr sourceAddress, int segmentId, ulong desti
                     var pageBlob = this.pageBlobDirectory.GetPageBlobClient(this.GetSegmentBlobName(segmentId));
 
                     // If segment size is -1 we use a default
-                    var size = this.segmentSize == -1 ? AzureStorageDevice.MAX_PAGEBLOB_SIZE : this.segmentSize;
+                    var size = this.segmentSize == -1 ? AzureStorageDevice.STARTING_DEFAULT_PAGEBLOB_SIZE : this.segmentSize;
```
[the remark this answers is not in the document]
sebastianburckhardt: Setting a max size for page blobs in settings makes sense to me. Will add that.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
davidmrdavid: I see. I think it would be good to write that in the inlined comment being referenced here. This also answers my question here: https://github.com/microsoft/durabletask-netherite/pull/329#discussion_r1378209082

PR #332, opened 2023-11-01
Revise partition startup

sebastianburckhardt:
While analyzing https://github.com/Azure/azure-functions-durable-extension/issues/2603 we noticed that in some cases, partition startup takes an extraordinary amount of time. This may be caused by hangs or other reasons.

To better diagnose this type of issue, and to make it can recover from indefinite hangs, this PR makes the following changes:

1. It adds a "waiting" transition prior to the "starting" transition. This allows us to differentiate between time spent waiting for the MaxConcurrentStarts semaphore, and the actual time spent starting up the partition.
2. I adds a timeout to the partition startup. This timeout is configurable in the settings. The default is 15 minutes for now.

### DIFF
--- a/src/DurableTask.Netherite/OrchestrationService/Partition.cs
+++ b/src/DurableTask.Netherite/OrchestrationService/Partition.cs
@@ -93,9 +93,6 @@ namespace DurableTask.Netherite
             EventTraceContext.Clear();
 
             this.ErrorHandler = errorHandler;
-
-            this.TraceHelper.TracePartitionProgress("Starting", ref this.LastTransition, this.CurrentTimeMs, "");
-
             errorHandler.Token.Register(() =>
             {
                 this.TraceHelper.TracePartitionProgress("Terminated", ref this.LastTransition, this.CurrentTimeMs, "");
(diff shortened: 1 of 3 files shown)


src/DurableTask.Netherite/OrchestrationService/Partition.cs
```diff
@@ -108,26 +105,37 @@ public async Task<(long, int)> CreateOrRestoreAsync(IPartitionErrorHandler error
                 }
 
             }, useSynchronizationContext: false);
-          
+
+            // before we start the partition, we have to acquire the MaxConcurrentStarts semaphore
+            // (to prevent a host from being overwhelmed by the simultaneous start of too many partitions)
+            this.TraceHelper.TracePartitionProgress("Waiting", ref this.LastTransition, this.CurrentTimeMs, "");
             await MaxConcurrentStarts.WaitAsync();
```
davidmrdavid, line 112: nit: can we print the capacity of `MaxConcurrentStarts` in this log somehow? If we're blocked after this log, knowing the number of competing concurrent starts would be helpful info.

src/DurableTask.Netherite/Util/PartitionTimeout.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        public void Dispose()
+        {      
+            // cancel the timeout task (if it has not already completed)
+            this.tokenSource.Cancel();
+
+            // dispose the token source after the timeout task has completed
+            this.timeoutTask.ContinueWith(_ => this.tokenSource.Dispose());
+        }
+    }
```
davidmrdavid, line 45: Isn't there a race condition here? What if the startup completes just a millisecond before the timeout task. If so, it's possible startup to complete and the timeout task to "continueWith" terminating the partition right after.

I suppose that may be 'ok' in the sense that if the startup takes 9.999 minutes, it might as well have timed out.

davidmrdavid (approved):
Some nits and suggestions but nothing blocking

src/DurableTask.Netherite/Util/PartitionTimeout.cs
sebastianburckhardt: It is true that that is possible.

Like you say, I think it is o.k. in this case.

Also, by design of the error handler, it doesn't make much of a difference whether the partition is terminated immediately before startup finishes or immediately after startup finishes.

sebastianburckhardt: That said, I found a way to fix this which is slightly cleaner.

sebastianburckhardt: I think I found a way to fix this so I will do that.

PR #335, opened 2023-12-18
Always release partition lease when a partition shuts down

sebastianburckhardt:
While analyzing some of traces showing high client latencies (see https://github.com/Azure/azure-functions-durable-extension/issues/2603) we discovered a problem: in some cases the partition lease is not released when the partition is terminated, because the request is cancelled. This causes an excessive delay before the partitition recovery, since the lease has to expire first. Waiting for the lease to expire can take up to 40s and delays all pending client requests to that partition.

This PR fixes this problem by not passing the termination cancellation token to the lease renewal (which caused the lease release to not happen).

### DIFF
--- a/src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
+++ b/src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
@@ -737,7 +737,10 @@ namespace DurableTask.Netherite.Faster
 
                     this.FaultInjector?.StorageAccess(this, "ReleaseLeaseAsync", "ReleaseLease", this.eventLogCommitBlob.Name);
 
-                    // we always release the lease here, whether the partition has already been terminated or not.
+                    // we must always release the lease here, whether the partition has already been terminated or not.
+                    // otherwise, the recovery of this partition (on this host or another host) has to wait for up
+                    // to 40s for the lease to expire. During this time, the partition is stalled (cannot process any work
+                    // items or client requests). In particular, client requests targeting this partition may be heavily delayed.
                     await this.leaseClient.ReleaseAsync(conditions: null, CancellationToken.None).ConfigureAwait(false);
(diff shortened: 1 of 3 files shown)


src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
```diff
@@ -736,17 +736,12 @@ public async Task MaintenanceLoopAsync()
                     this.TraceHelper.LeaseProgress("Releasing lease");
 
                     this.FaultInjector?.StorageAccess(this, "ReleaseLeaseAsync", "ReleaseLease", this.eventLogCommitBlob.Name);
-                    await this.leaseClient.ReleaseAsync(null, this.PartitionErrorHandler.Token).ConfigureAwait(false);
+
+                    // we always release the lease here, whether the partition has already been terminated or not.
+                    await this.leaseClient.ReleaseAsync(conditions: null, CancellationToken.None).ConfigureAwait(false);
+
```
davidmrdavid, line 742: nit:
```suggestion

                    // we always release the lease here, whether the partition has already been terminated or not.
                    // otherwise, a pending client request for that partition will be delayed until the lease expires
                    await this.leaseClient.ReleaseAsync(conditions: null, CancellationToken.None).ConfigureAwait(false);

```

davidmrdavid (approved):

cgillum (approved):

sebastianburckhardt:
Added two more changes (because of related issues observed when testing this PR).
These are small "tuning" changes that may help a little when dealing with OOM, for example.

- recognize another transient storage exception (so it can be retried instead of failing the partition immediately)
- change the default setting so partitions do NOT take checkpoints when shutting down (taking many checkpoints at once is a typical cause for OOM)
- do not release lease if it is already lost

davidmrdavid (approved):

PR #341, opened 2024-01-10
Add RecoveryTester feature

sebastianburckhardt:
This PR adds a new feature for testing the recovery of partitions without modifying any files in the storage account.

The feature is available in the form of a new option for the `PartitionManagement` setting:

```
   "PartitionManagement": "RecoveryTester",
   "PartitionManagementParameters": "7",
```

(specifying the partition is optional, if none is specified, all partitions are tested).

As part of this PR, I also
- refactored how partition managers are selected based on the PartitionManagement setting. This is in anticipation of adding a static partition manager setting, to replace the 'Scripted' setting.
- removed the 'Scripted' setting since this was meant for internal development only and I have not been using it and it for years, so it does not make sense to maintain that code.


src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationServiceSettings.cs
```diff
@@ -91,6 +91,11 @@ public class NetheriteOrchestrationServiceSettings
         [JsonConverter(typeof(StringEnumConverter))]
         public PartitionManagementOptions PartitionManagement { get; set; } = PartitionManagementOptions.EventProcessorHost;
 
+        /// <summary>
+        /// Additional parameters for the partition management, if necessary
+        /// </summary>
+        public string PartitionManagementParameters { get; set; } = null;
+
```
davidmrdavid, line 98: From what I gather, this setting is rather dynamic, is it not? It can specify the partition ID if using the RecoverTester, but in the future it could specify other stuff (that's why the type is `string`), right?

I would prefer a more constrained/type-safe approach here, as I think parsing strings can get hairy over time. Could be create a proper options object?

src/DurableTask.Netherite/OrchestrationService/PartitionManagementOptions.cs
```diff
@@ -18,13 +18,20 @@ public enum PartitionManagementOptions
...
+        /// Use the event processor host implementation provided by the EventHubs client library. This dynamically
+        /// balances the partitions across all connected hosts.
         /// </summary>
         EventProcessorHost,
 
         /// <summary>
         /// Follow a predefined partition management script. This is meant to be used for testing and benchmarking scenarios.
+        /// This was an internal feature and is no longer supported.
         /// </summary>
         Scripted,
```
davidmrdavid, line 30: nit: I'm curious if we can mark this as obsolete to prevent future usage.

```suggestion
        /// <summary>
        /// Follow a predefined partition management script. This is meant to be used for testing and benchmarking scenarios.
        /// This was an internal feature and is no longer supported.
        /// </summary>
         [Obsolete]
        Scripted,
```

src/DurableTask.Netherite/OrchestrationService/PartitionManagementOptions.cs
```diff
@@ -18,13 +18,20 @@ public enum PartitionManagementOptions
...
+
+        /// <summary>
+        /// Test the recovery without modifying storage. Useful for diagnosing recovery problems. Not functional as an orchestration service!
+        /// </summary>
+        RecoveryTester,
```
davidmrdavid, line 35: nit: I wonder if adding `[Obsolete]` here would also prevent usage in prod.

```suggestion
        /// <summary>
        /// Test the recovery without modifying storage. Useful for diagnosing recovery problems. Not functional as an orchestration service! Marked as 'Obsolete' to prevent usage in production.
        /// </summary>
        [Obsolote]
        RecoveryTester,
```

src/DurableTask.Netherite/StorageLayer/Faster/PartitionStorage.cs
```diff
@@ -252,21 +260,24 @@ void Check(string workerName, TimeSpan? busySince)
...
-            // observe exceptions if the clean shutdown is not working correctly
-            await this.TerminationWrapper(t1);
-            await this.TerminationWrapper(t2);
-
-            // if the the settings indicate we want to take a final checkpoint, do so now.
-            if (takeFinalCheckpoint)
+            if (!this.blobManager.ReadonlyMode)
             {
-                this.TraceHelper.FasterProgress("Writing final checkpoint");
-                await this.TerminationWrapper(this.storeWorker.TakeFullCheckpointAsync("final checkpoint").AsTask());
```
davidmrdavid, line 269: If we didn't have this if-statement, would our `failIfReadOnly` parameter still help us prevent writes? No, right?

davidmrdavid (commented):
A few first impressions.
Also - did you get a chance to test that read-only mode is in fact leaving a storage account as-is? We may want to try it out on a throwaway account first

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationServiceSettings.cs
sebastianburckhardt: Right, this is anticipating future, alternate partition managers where the parameter can be something much different (e.g. it could be an entire json object).

> Could be create a proper options object?

Don't mind the idea but I am totally unsure whether/how that can work in the context of all the stuff the functions host does with host.json (e.g. magic environment variable settings and the like). So I was just taking the easy route.
If you have experience with it, let me know.

src/DurableTask.Netherite/OrchestrationService/PartitionManagementOptions.cs
sebastianburckhardt: marking it as obsolete may violates major-minor version change rules.

src/DurableTask.Netherite/OrchestrationService/PartitionManagementOptions.cs
sebastianburckhardt: I prefer not to since it is not actually obsolete. 

I don't think I have ever seen anybody run into trouble because they are using undocumented configuration settings. I think they know better. There are a ton of such settings, and they are not currently marked obsolete.

src/DurableTask.Netherite/StorageLayer/Faster/PartitionStorage.cs
sebastianburckhardt: No. Inserting this guard was the only way to prevent writes. 

THere is no systematic way to prevent writes as far as I know. Also, systematic prevention would be wrong since there are still some accesses that modify storage (e.g. the leases, and the load publishing).

I am quite sure all writes to application data (partition state) are going through those two calls in PartitionStorage, so I think that is good enough.

sebastianburckhardt:
> Also - did you get a chance to test that read-only mode is in fact leaving a storage account as-is? 

I did some testing but did not actually look at storage afterwards to confirm it is unmodified. Can do that.

src/DurableTask.Netherite/OrchestrationService/PartitionManagementOptions.cs
davidmrdavid: Fair enough, resolving.

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationServiceSettings.cs
davidmrdavid: Gotcha. I do have some experience here (but not extensive) and I wanna say that the parsing will of `host.json` is an issue for the DF Extension, so I think that's a problem we can solve there instead of here. In general, it's not that magical :) .

In this case, yeah I think the ""right thing to do"" is to create a partitionManagement options class or interface with a subclass/implementation specific for the recoveryTester. That should suffice for us at this layer. I can solve the problem of parsing at the DF Extension layer :)

sebastianburckhardt: Got it. But for the sake of using our time wisely, I think it is still an o.k. solution to just use a string here.

sebastianburckhardt:
@davidmrdavid, are you o.k. with merging this PR now? I think the recovery feature has proved useful and there seems little risk with putting this in the main branch.

PR #344, opened 2024-02-07
Upgrade FASTER dependency to v2.6.1

bachuv:
This PR includes the following changes:

- [ ] Upgrade FASTER dependency to v2.6.1
- [ ] Upgrade Netherite version to 2.0.0-preview
- [ ] Add validation that customers are not using Netherite V2 with a task hub from Netherite V1 and vice versa
- [ ] Add validation when serializing/deserializing so we can more easily pinpoint data corruption issues

### DIFF
--- a/src/DurableTask.Netherite/DurableTask.Netherite.csproj
+++ b/src/DurableTask.Netherite/DurableTask.Netherite.csproj
@@ -24,11 +24,11 @@
 
   <!-- Version settings: https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/ -->
   <PropertyGroup>
-    <MajorVersion>1</MajorVersion>
-	<MinorVersion>4</MinorVersion>
-	<PatchVersion>1</PatchVersion>
+    <MajorVersion>2</MajorVersion>
+	<MinorVersion>0</MinorVersion>
+	<PatchVersion>0</PatchVersion>
 	<VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
-    <VersionSuffix></VersionSuffix>
+    <VersionSuffix>preview</VersionSuffix>
     <AssemblyVersion>$(MajorVersion).0.0.0</AssemblyVersion>
     <BuildSuffix Condition="'$(GITHUB_RUN_NUMBER)' != ''">.$(GITHUB_RUN_NUMBER)</BuildSuffix>
     <FileVersion>$(VersionPrefix)$(BuildSuffix)</FileVersion>
@@ -56,7 +56,7 @@
     <PackageReference Include="Microsoft.Azure.EventHubs.Processor" Version="4.3.2" />
     <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="11.2.3" />
     <PackageReference Include="Azure.Storage.Blobs" Version="12.16.0" />
-	<PackageReference Include="Microsoft.FASTER.Core" Version="2.0.23" />
+	<PackageReference Include="Microsoft.FASTER.Core" Version="2.6.1" />
     <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.15.1" />
 	<PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
--- a/src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
+++ b/src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
@@ -121,7 +121,6 @@ namespace DurableTask.Netherite.Faster
                 MutableFraction = tuningParameters?.StoreLogMutableFraction ?? 0.9,
                 SegmentSizeBits = segmentSizeBits,
                 PreallocateLog = false,
-                ReadFlags = ReadFlags.None,
                 ReadCacheSettings = null, // no read cache
                 MemorySizeBits = memorySizeBits,
             };
@@ -873,7 +872,7 @@ namespace DurableTask.Netherite.Faster
 
         #region ILogCommitManager
 
-        void ILogCommitManager.Commit(long beginAddress, long untilAddress, byte[] commitMetadata, long commitNum)
+        void ILogCommitManager.Commit(long beginAddress, long untilAddress, byte[] commitMetadata, long commitNum, bool forceWriteMetadata)
         {
             try
             {
@@ -1458,6 +1457,11 @@ namespace DurableTask.Netherite.Faster
                         this.CheckpointInfoETag = response.Value.ETag;
                     });
             }
+        }
+
+        public void CheckpointVersionShift(long oldVersion, long newVersion)
+        {
+            throw new NotImplementedException();
         }
     }
 }
--- a/src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/LocalFileCheckpointManager.cs
+++ b/src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/LocalFileCheckpointManager.cs
@@ -91,5 +91,10 @@ namespace DurableTask.Netherite.Faster
 
         void IDisposable.Dispose()
             => this.localCheckpointManager.Dispose();
+
+        public void CheckpointVersionShift(long oldVersion, long newVersion)
+        {
+            throw new NotImplementedException();
+        }
     }
 }
--- a/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
+++ b/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
@@ -186,7 +186,7 @@ namespace DurableTask.Netherite.Faster
         ClientSession<Key, Value, EffectTracker, Output, object, IFunctions<Key, Value, EffectTracker, Output, object>> CreateASession(string id, bool isScan)
         {
             var functions = new Functions(this.partition, this, this.cacheTracker, isScan);
-            return this.fht.NewSession(functions, id, readFlags: (isScan ? ReadFlags.None : ReadFlags.CopyReadsToTail));
+            return this.fht.NewSession(functions, id);
         }
 
         public IDisposable TrackTemporarySession(ClientSession<Key, Value, EffectTracker, Output, object, IFunctions<Key, Value, EffectTracker, Output, object>> session)
--- a/test/DurableTask.Netherite.Tests/DurableTask.Netherite.Tests.csproj
+++ b/test/DurableTask.Netherite.Tests/DurableTask.Netherite.Tests.csproj
@@ -9,7 +9,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="6.0.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.2" />
     <PackageReference Include="xunit" Version="2.4.2" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.5">


src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
```diff
@@ -873,7 +872,7 @@ async Task DeleteCheckpointDirectory(IEnumerable<string> blobsToDelete)
 
         #region ILogCommitManager
 
-        void ILogCommitManager.Commit(long beginAddress, long untilAddress, byte[] commitMetadata, long commitNum)
+        void ILogCommitManager.Commit(long beginAddress, long untilAddress, byte[] commitMetadata, long commitNum, bool forceWriteMetadata)
```
bachuv, line 875: How should we use `forceWriteMetadata` in this method?

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
```diff
@@ -1459,5 +1458,10 @@ internal async Task FinalizeCheckpointCompletedAsync()
                     });
             }
         }
+
+        public void CheckpointVersionShift(long oldVersion, long newVersion)
+        {
+            throw new NotImplementedException();
+        }
```
bachuv, line 1465: Any pointers on how to implement this method? Where should we update the version?

src/DurableTask.Netherite/StorageLayer/Faster/FasterStorageProvider.cs
```diff
@@ -134,6 +134,8 @@ async Task<bool> IStorageLayer.CreateTaskhubIfNotExistsAsync()
                 this.traceHelper.TraceProgress($"Using existing blob container at {this.cloudBlobContainer.Result.Uri}");
             }
 
+            // TODO: Add validation to check if customer is using a Netherite V1 task hub
```
bachuv, line 137: Is this the correct place to add validation to make sure customers are not using Netherite V1 task hubs with Netherite V2 and vice versa? Should we check the `StorageFormat` property for this check? If so, what values can we expect for `StorageFormat`?

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
davidmrdavid: In case it helps, this seems defined here: https://github.com/microsoft/FASTER/blob/0116754bcf0857e614e2b7ea042b8db563cb26cc/cs/src/core/FasterLog/ILogCommitManager.cs#L23

and more context may be found here: https://github.com/microsoft/FASTER/blob/0116754bcf0857e614e2b7ea042b8db563cb26cc/cs/src/core/FasterLog/FasterLog.cs#L2077-L2083

Now that still doesn't make the answer obvious to me, but it gives me some clues. Perhaps this is something we can ask the FASTER folks about

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
davidmrdavid: Also worth checking this w/ FASTER team, however, from looking at the source: https://github.com/microsoft/FASTER/blob/0116754bcf0857e614e2b7ea042b8db563cb26cc/cs/src/core/Index/Recovery/ICheckpointManager.cs#L61-L66

it appears this is just a callback for us, that we may ignore. I.e, we could use this callback to verbose log that the checkpoint version has increased, but it doesn't have to do real work. Again, just this is just a best guess though :)

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
```diff
@@ -186,7 +186,7 @@ void Shutdown()
         ClientSession<Key, Value, EffectTracker, Output, object, IFunctions<Key, Value, EffectTracker, Output, object>> CreateASession(string id, bool isScan)
         {
             var functions = new Functions(this.partition, this, this.cacheTracker, isScan);
-            return this.fht.NewSession(functions, id, readFlags: (isScan ? ReadFlags.None : ReadFlags.CopyReadsToTail));
```
davidmrdavid, line 189: so how did this get dropped?

src/DurableTask.Netherite/StorageLayer/Faster/FasterStorageProvider.cs
davidmrdavid: I feel it might be better to do this validation closer to the `return` statement of this method. In particular, I'd do it in (or after) the `catch` block that reads:

```C#
            catch (StorageException e) when (BlobUtils.BlobAlreadyExists(e))
            {
                // taskhub already exists, possibly because a different node created it faster
                this.traceHelper.TraceProgress("Confirmed existing taskhub");
                return false;
            }
```

In that catch block, we know the taskhub exists. So it may be possible to read it and check some property of it.

Not sure what property to check to determine if the taskhub was created w/ Netherite V1. I'd place a debugger to see what gets loaded at runtime to try to infer what data there can be augmented to include the Netherite version.

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
bachuv: `NewSession()`doesn't take an argument named `readFlags` after upgrading the FASTER version so it seems like there's an API change here. There's a `ReadCopyOptions` parameter though so maybe this information should be sent through that. This would be good to clarify with the FASTER team.

TedHartMS: Correct--we've improved the control over copy source/destination in later versions. Here you probably want:
   isScan ? default : new {CopyFrom = ReadCopyFrom.Device, CopyTo = ReadCopyFrom.MainLog};

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
sebastianburckhardt: I believe it is safe to treat this callback as a no-op (which is what David is suggesting) since the default commit manager in FASTER (DeviceLogCommitCheckpointManager) also treats this as a no-op, so it can't really be important.

src/DurableTask.Netherite/StorageLayer/Faster/FasterStorageProvider.cs
sebastianburckhardt: The validation check already exists. All you have to do is add another line at the end of the `StorageFormatVersion` variable at https://github.com/microsoft/durabletask-netherite/blob/60c7d39e03ec35990f2115efc07ca1d09ceb56e1/src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs#L152C8-L159C1

We are not sure yet that we need it though. Still trying to get this to be a non-breaking change.

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
sebastianburckhardt: To keep it consistent with the current Netherite design I think we should copy from the immutable region also. So I would suggest something like this:

```csharp
            ReadCopyOptions readCopyOptions = isScan 
                ? new ReadCopyOptions(ReadCopyFrom.None, ReadCopyTo.None) 
                : new ReadCopyOptions(ReadCopyFrom.AllImmutable, ReadCopyTo.MainLog);
            
            return this.fht.NewSession(functions, id, default, readCopyOptions);
```

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/LocalFileCheckpointManager.cs
```diff
@@ -91,5 +91,10 @@ public void OnRecovery(Guid indexToken, Guid logToken)
 
         void IDisposable.Dispose()
             => this.localCheckpointManager.Dispose();
+
+        public void CheckpointVersionShift(long oldVersion, long newVersion)
+        {
+            throw new NotImplementedException();
```
sebastianburckhardt, line 97:
```suggestion
            // no-op
```

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
```diff
@@ -121,7 +121,6 @@ public LogSettings GetStoreLogSettings(bool useSeparatePageBlobStorage, long upp
                 MutableFraction = tuningParameters?.StoreLogMutableFraction ?? 0.9,
                 SegmentSizeBits = segmentSizeBits,
                 PreallocateLog = false,
-                ReadFlags = ReadFlags.None,
                 ReadCacheSettings = null, // no read cache
```
sebastianburckhardt, line 124:
```suggestion
                ReadCopyOptions = default, // is overriden by the per-session configuration
                ReadCacheSettings = null, // no read cache
```

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
sebastianburckhardt: After looking at where this was  introduced in FASTER (https://github.com/microsoft/FASTER/pull/856) it is pretty clear that this fixes a problem that was introduced with "fast commit" where the commit is not always fully committed.

In our implementation we *always fully commit* anyway so we can ignore this argument.

sebastianburckhardt (changes requested):
Overall, looks pretty good. Left some comments.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobManager.cs
TedHartMS: That's true as far as I can see; it's called during the state machine processing but nothing depends on the result. So just treat as a no-op

src/DurableTask.Netherite/StorageLayer/Faster/FasterStorageProvider.cs
sebastianburckhardt: Based on our discussion last week, once we get PR https://github.com/microsoft/FASTER/pull/896 merged and published, we should be able to keep this a non-breaking change so we don't need the version check.

src/DurableTask.Netherite/DurableTask.Netherite.csproj
```diff
@@ -56,7 +56,7 @@
     <PackageReference Include="Microsoft.Azure.EventHubs.Processor" Version="4.3.2" />
     <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="11.2.3" />
     <PackageReference Include="Azure.Storage.Blobs" Version="12.16.0" />
-	<PackageReference Include="Microsoft.FASTER.Core" Version="2.0.23" />
+	<PackageReference Include="Microsoft.FASTER.Core" Version="2.6.1" />
```
sebastianburckhardt, line 59: As soon as it is available we should change this to 2.6.2.

src/DurableTask.Netherite/StorageLayer/Faster/FasterStorageProvider.cs
sebastianburckhardt, line 137:
```suggestion
```

sebastianburckhardt (commented):
Looks good, some very minor changes, and waiting for the updated FASTER.

Then we can merge I think, but before releasing, we will still need to run tests to validate (a) users can keep using their existing task hubs without breaking changes, and (b) that this new version of FASTER can run our workloads without new error popping up.

src/DurableTask.Netherite/DurableTask.Netherite.csproj
sebastianburckhardt, line 59:
```suggestion
	<PackageReference Include="Microsoft.FASTER.Core" Version="2.6.2" />
```

FASTER v2.6.2 has been released.

sebastianburckhardt (approved):

sebastianburckhardt:
Starting to run some tests now.

sebastianburckhardt:
There appears to be a problem with not being able to load v7.0.0 of Microsoft.Extensions.Logging.

I opened a PR to modify FASTER to remain compatible.

sebastianburckhardt:
I did a quick validation that it is possible to open old taskhubs after updating.

*However note* that the reverse is not true. Once you have opened a task hub with this version, you cannot go back to an older version! We may need to include a warning?

davidmrdavid:
@sebastianburckhardt: I don't think we can detect this scenario in Netherite, can we? From looking at the diff, it seems the loading of  previous FASTER format is happening at the FASTER level, and our code is oblivious to it, is it not?

Assuming this is the case, I'm personally 'ok' with it. However, I think this should be a minor version increase, not a patch version, since we're updating a dependency and the minor version allows us to signal to users that they should pay extra attention to the release notes.

davidmrdavid (changes requested):
nit but important: let's please make this a minor version release :) . We're upgrading a dependency and there's implications in terms of backwards compat with previous FASTER storage formats

sebastianburckhardt:
@bachuv, I think you can merge this once @davidmrdavid approves.

We should still run more tests before releasing 1.5.0 but it makes sense to test the entire release, not just this PR.

> let's please make this a minor version release

I have added a PR #368 to update the version to 1.5.0, so it does not need to be part of this PR.

davidmrdavid (approved):
🚢 it!

bachuv:
> @bachuv, I think you can merge this once @davidmrdavid approves.

Okay, I'll merge it once the CI passes.

sebastianburckhardt:
Hmm, those CI failures look somewhat concerning given that FASTER has undergone a lot of updates (and therefore possibly contains new bugs). I will take a look at those failures.

sebastianburckhardt:
The CI failures were bugs in FASTER which were fixed in https://github.com/microsoft/FASTER/pull/911. We should be able to move forward as soon as FASTER v2.6.4 is released.

PR #347, opened 2024-02-12
fix EH batch load cancellation issue

sebastianburckhardt:
Testing revealed issues where partitions continuously fail because of supposed "out-of-order" packets. What is actually happening is that the redelivery of previously delivered packets is sometimes not working correctly after a partition terminates.

I tracked this down to the fact that the event hubs transport may be inadvertently cancelling storage requests that retrieve the batch contents when the partition is terminated, which is incorrect (loading the batch content is not tied to the partition state and should be succeeding even if the partition is recycled). This means some packets are skipped and never delivered. 

This PR fixes the problem by using the correct cancellation token.

It also adds a condition to the message to check whether any cancellations are o.k. (so that inadvertent cancellations are reported as errors).

### DIFF
--- a/src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsProcessor.cs
+++ b/src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsProcessor.cs
@@ -437,7 +437,7 @@ namespace DurableTask.Netherite.EventHubsTransport
                     // iterators do not support ref arguments, so we use a simple wrapper class to work around this limitation
                     MutableLong nextPacketToReceive = new MutableLong() { Value = current.NextPacketToReceive.seqNo };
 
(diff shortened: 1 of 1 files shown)


src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsProcessor.cs
```diff
@@ -489,7 +489,7 @@ await foreach ((EventData eventData, PartitionEvent[] events, long seqNo) in thi
 
                 await this.SaveEventHubsReceiverCheckpoint(context, 600000);
             }
-            catch (OperationCanceledException)
+            catch (OperationCanceledException) when (this.shutdownToken.IsCancellationRequested)
```
davidmrdavid, line 492:
```suggestion
            catch (OperationCanceledException) when (this.shutdownToken.IsCancellationRequested) // we should only ignore these exceptions during VM shutdowns. See :  https://github.com/microsoft/durabletask-netherite/pull/347
```

davidmrdavid (approved):
approved w/ nit

PR #349, opened 2024-02-13
Improve error handling for FASTER IO completion callbacks

sebastianburckhardt:
Two changes:
- wrap FASTER callbacks so that exceptions are noticed 
- release semaphore prior to calling back into FASTER (to avoid losing the semaphore should FASTER hang)


src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
```diff
@@ -243,15 +243,31 @@ void CancelAllRequests()
...
                     }
-                    request.Callback(uint.MaxValue, request.NumBytes, request.Context);
+
+                    try
+                    {
+                        request.Callback(uint.MaxValue, request.NumBytes, request.Context);
+                    }
+                    catch (Exception ex)
+                    {
+                        this.BlobManager.StorageTracer?.FasterStorageError($"{nameof(CancelAllRequests)} for access id={id} failed during FASTER completion callback", ex);
```
davidmrdavid, line 253: would it make sense to re-throw here? Aren't we swallowing the exception

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
```diff
@@ -380,23 +396,9 @@ public override unsafe void ReadAsync(int segmentId, ulong sourceAddress, IntPtr
...
-                          }
-                          else
-                          {
-                              this.BlobManager?.StorageTracer?.FasterStorageProgress($"StorageOpReturned AzureStorageDevice.ReadAsync id={id}");
-                              request.Callback(0, request.NumBytes, request.Context);
-                          }
-                      }
-                  }, TaskContinuationOptions.ExecuteSynchronously);
+            // we are not awaiting this task because it uses FASTER's callback mechanism
+            // when the access is completed.
```
davidmrdavid, line 400: can't immediately see why it's safe to remove this logic. Mind elaborating? Perhaps it's related to your comment that this task isn't awaited, but I did not fully get that either.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
```diff
@@ -555,45 +558,62 @@ async Task ReadFromBlobAsync(UnmanagedMemoryStream stream, BlobUtilsV12.PageBlob
...
+                IsFaulted = true;
+
+                // details about failed storage accesses are already traced by the error handler, but we still create another trace here
+                // to help us debug situations where exceptions are thrown in other places, and to clarify the execution path
+                this.BlobManager.StorageTracer?.FasterStorageError($"{nameof(ReadFromBlobAsync)} id={id} encountered exception", e);
+            }
+            finally
+            {
+                stream.Dispose();
+            }
```
davidmrdavid, line 572: Is there a reason to explicitly dispose the `stream` instead of continue to use the `using(stream)` statement, as before?

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
```diff
@@ -603,14 +623,58 @@ async Task WriteToBlobAsync(BlobEntry blobEntry, IntPtr sourceAddress, long dest
...
+                // details about failed storage accesses are already traced by the error handler, but we still create another trace here
+                // to help us debug situations where exceptions are thrown in other places, and to clarify the execution path
+                this.BlobManager.StorageTracer?.FasterStorageError($"{nameof(WriteToBlobAsync)} id={id} encountered exception", e);
             }
+            finally
+            {
+                if (this.underLease)
+                {
+                    this.SingleWriterSemaphore.Release();
+                }
```
davidmrdavid, line 652: nit: please add comment on why this is important to do at this point

davidmrdavid: Also, do we not have a forceful semaphore release if it hangs for too long?

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
```diff
@@ -603,14 +623,58 @@ async Task WriteToBlobAsync(BlobEntry blobEntry, IntPtr sourceAddress, long dest
...
+                    {
+                        this.BlobManager?.StorageTracer?.FasterStorageProgress($"StorageOpReturned AzureStorageDevice.WriteAsync id={id}");
+                        request.Callback(0, request.NumBytes, request.Context);
+                    }
+                    this.BlobManager?.StorageTracer?.FasterStorageProgress($"AzureStorageDevice.WriteAsync FASTER callback completed id={id}");
+                }
+                catch (Exception e)
+                {
+                    this.BlobManager.StorageTracer?.FasterStorageError($"{nameof(WriteToBlobAsync)} id={id} failed during FASTER completion callback", e);
+                }
```
davidmrdavid, line 674: should we re-throw?

davidmrdavid (commented):
Some minor questions

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
sebastianburckhardt: There is no one to catch this (it is an un-awaited task) and having exceptions in un-awaited tasks can be harmful as far as I know.

sebastianburckhardt: There is no one to meaningfully catch exceptions here. And the partition has already been terminated (that's why the requests are being cancelled). So it is just about tracing at this point.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
sebastianburckhardt: Yeah, I did not like to have both `try` and `using` - too much nesting.

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
sebastianburckhardt: There is no one to catch this. (see comment at bottom of method)

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/AzureStorageDevice.cs
sebastianburckhardt: > lease add comment on why this is important to do at this point
ok

 > do we not have a forceful semaphore release if it hangs for too long?
the hang check is somewhere else. We don't need it here since the entire object (incl the semaphore) gets disposed when the partition is terminated. Also, have not seen Azure Storage hanging recently.

davidmrdavid (approved):

sebastianburckhardt:
Sorry for the churn but I did a substantial refactor of this. The error handling path is simplified compared to before. A significant change is that I am no longer calling cancellation callbacks into FASTER at all. I noticed that those cancellation callbacks were causing a lot of exceptions and it turns out they are not needed.

davidmrdavid (approved):

PR #351, opened 2024-02-13
Fix hanging blob creation

sebastianburckhardt:
We observed an unnaturally large number of errors like "storage operation id=2438 has exceeded the time limit 00:01:30" that terminate the partition.

After analysis we found that this is caused by situations where a blob creation simultaneously succeeds (i.e. blob was created) and times out. Our retry path was not correctly detecting that the blob was already created.

This PR:
- fixes a missing check in the retry condition to avoid unlimited retries
- adds the id into the traces for the blob creation for easier diagnostics
- if the blob already exists, loads the e-tag, and skips the next attempt

### DIFF
--- a/src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobEntry.cs
+++ b/src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobEntry.cs
@@ -90,14 +90,8 @@ namespace DurableTask.Netherite.Faster
                 },
                 async () =>
                 {
-                    try
-                    {
-                        var response = await pageBlob.Default.GetPropertiesAsync();
-                        this.ETag = response.Value.ETag;
-                    }
-                    catch (Azure.RequestFailedException ex) when (BlobUtilsV12.BlobDoesNotExist(ex))
-                    {
-                    }
+                    var response = await pageBlob.Default.GetPropertiesAsync();
+                    this.ETag = response.Value.ETag;
                 });
 
             // At this point the blob is fully created. After this line all consequent writers will write immediately. We just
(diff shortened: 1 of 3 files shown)


src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobEntry.cs
```diff
@@ -85,8 +90,14 @@ public async Task CreateAsync(long size, BlobUtilsV12.PageBlobClients pageBlob)
...
-                    var response = await pageBlob.Default.GetPropertiesAsync();
-                    this.ETag = response.Value.ETag;
+                    try
+                    {
+                        var response = await pageBlob.Default.GetPropertiesAsync();
+                        this.ETag = response.Value.ETag;
+                    }
+                    catch (Azure.RequestFailedException ex) when (BlobUtilsV12.BlobDoesNotExist(ex))
+                    {
+                    }
```
davidmrdavid, line 100: So why are we swallowing this exception?

davidmrdavid (commented):
one clarifying question

src/DurableTask.Netherite/StorageLayer/Faster/AzureBlobs/BlobEntry.cs
davidmrdavid: I figured it would be good to at least log it, assuming the number of retries isn't all that large.

sebastianburckhardt: This code gets called if an attempt to create the blob _appears_ to have failed (but may actually have succeeded). In that situation we don't know if the blob does or does not exist. If creation was actually successful, we load the e-tag which then skips creation on the next attempt.

sebastianburckhardt: After looking at this again, there is actually no reason to swallow the exception because it is not actually expected.

(This code is only called after a previous creation failed with an e-tag exception, so at that point the blob really should exist).

I am therefore removing this try-catch.

Issue #352, opened 2024-02-14
NetheriteOrchestrationService.StartAsync throws cached exception on retry if first call fails


cgillum:
The following was observed in a customer support case.

1. The function app starts up and Netherite throws an exception because of a transient DNS issue when connecting to Event Hubs.
1. The Functions host handles the failure, and tries to restart the host startup process.
1. NetheriteOrchestrationService.StartAsync gets called again with the same NetheriteOrchestrationService object instance, but the same exception gets thrown. This time, however, we don't log it. The DNS error only shows up in the FunctionsLogs - not the DurableFunctionsEvents log, indicating that we're not actually running the startup logic a second time.
1. This continues indefinitely.
1. The problem stops only when the customer forcefully restarts the function app

As discussed, this appears to be a caching issue in NetheriteOrchestrationService. The Azure Functions Host assumes that we can retry startup failures, which is why we see this behavior in production. However, NetheriteOrchestrationService.StartAsync doesn't appear to support retrying.

In order to be compatible with the Azure Functions host retry logic, we should change this behavior.

Context from @sebastianburckhardt 

> ...this breaks down into two internal tasks (for starting client and for starting workers) which are being cached. I believe the reason for this caching is to deal with early and/or concurrent client calls and make exceptions in the various startup tasks visible to the application. I agree this is complicated and I would love something simpler but it has been tricky last time I tried because the environment does not follow what would be easy to implement (wait for successful StartAsync before calling any client operations)

> It would probably not be all that difficult to just clear all the cached failed transitions when StartAsync() is retried

bachuv:
Reopening this issue since we received a customer support case about the same problem and don't see the cache clearing.

PR #376, opened 2024-04-08
Clear cached exception when StartAsync is invoked

davidmrdavid:
Fix: https://github.com/microsoft/durabletask-netherite/issues/352

We know that `StartAsync` can be called by the Functions Host several times during start up, and therefore it should be possible to call this method more than once an obtain a different result. Today, that's not possible since we cache the exception.

This PR is a naive first attempt at addressing this: it discard the cached exception whenever "StartAsync" is invoked. This means that "early" attempts at creating clients will fail (as they'll reference the cached exception, but the Host should be clear to make progress on subsequent calls to `StartAsync`.

Open questions:
(1) Is there more cached state to clear? For example - `this.checkedClient` or `this.client`?
(2) What is the lifecycle of StartAsync? Is it guaranteed that the host has called `StopAsync` before it calls `StartAsync` again, or not? If this is not guaranteed, then we'll need to consider a few more race conditions, such as the start-up of cthe client and worker threads.

### DIFF
--- a/src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
+++ b/src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
@@ -290,7 +290,7 @@ namespace DurableTask.Netherite
         /// <inheritdoc />
         Task IOrchestrationService.StartAsync()
         {
-            // We discard exception state, as Functions Host may retry the StartAsync operation.
+            // We discard exception state, as hosts (like Azure Functions) may retry the StartAsync operation multiple times.
             // See: https://github.com/microsoft/durabletask-netherite/issues/352
             this.startupException = null;
 
--- a/src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
+++ b/src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
@@ -400,12 +400,17 @@ namespace DurableTask.Netherite
             {
                 this.TraceHelper.TraceProgress("Starting Client");
 
+                this.serviceShutdownSource ??= new CancellationTokenSource();
+
                 if (this.Settings.TestHooks != null)
                 {
                     this.TraceHelper.TraceProgress(this.Settings.TestHooks.ToString());
-                }
 
-                this.serviceShutdownSource ??= new CancellationTokenSource();
+                    if (this.Settings.TestHooks.FaultInjectionActive)
+                    {
+                        this.Settings.TestHooks.FaultInjector.StartClient();
+                    }
+                }
 
                 this.TaskhubParameters = await this.transport.StartAsync();
                 (this.ContainerName, this.PathPrefix) = this.storage.GetTaskhubPathPrefix(this.TaskhubParameters);
--- a/src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
+++ b/src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
@@ -408,7 +408,7 @@ namespace DurableTask.Netherite
 
                     if (this.Settings.TestHooks.FaultInjectionActive)
                     {
-                        this.Settings.TestHooks.FaultInjector.StartClient();
+                        this.Settings.TestHooks.FaultInjector.ClientStartup();
                     }
                 }
 
--- a/src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
+++ b/src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
@@ -359,23 +359,30 @@ namespace DurableTask.Netherite
             }
         }
 
-        static void TryTransition<T>(ref Task<T> transition, Task<T> lastTransition, Func<Task<T>> nextTransition)
+        static void TryTransition<T>(ref Task<T> currentTransition, Task<T> lastTransition, Func<Task<T>> nextTransition)
         {
-            var greenLight = new TaskCompletionSource<bool>();
+            // some other thread could be trying to do a transition at the same time, creating a race.
+            var waitForRaceToBeDetermined = new TaskCompletionSource<bool>();
+            // wrap the transition in a async function that waits for the race to be won before executing the transition
             var task = conditionalTransition();
-            var interlockedResult = Interlocked.CompareExchange<Task<T>>(ref transition, task, lastTransition);
-            bool success = interlockedResult == lastTransition;
-            greenLight.SetResult(success);
+
+            // to resolve any potential races, we use an interlocked operation. This operation replaces the currentTransition
+            // ONLY if the lastTransition matches the lastTransition that was observed earlier. Otherwise it does nothing.
+            var interlockedResult = Interlocked.CompareExchange<Task<T>>(ref currentTransition, task, lastTransition);
+            bool interlockedWasEffective = interlockedResult == lastTransition; // interlocked exchange returns the value it read
+            waitForRaceToBeDetermined.SetResult(interlockedWasEffective);
 
             async Task<T> conditionalTransition()
             {
-                if (await greenLight.Task)
+                if (await waitForRaceToBeDetermined.Task)
                 {
+                    // this thread won the race so we are now executing the transition.
                     return await nextTransition();
                 }
                 else
                 {
-                    return default; // this thread lost the race so the task does nothing
+                    // this thread lost the race so the task does nothing.
+                    return default; 
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
```diff
@@ -290,6 +290,10 @@ async Task IOrchestrationService.DeleteAsync()
         /// <inheritdoc />
         Task IOrchestrationService.StartAsync()
         {
+            // We discard exception state, as Functions Host may retry the StartAsync operation.
```
cgillum, line 293: I suggest mentioning the Functions Host as an example of a host that retries StartAsync failures, since the problem we're fixing is not necessarily specific to Azure Functions (nor is this Netherite code).

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
```diff
@@ -290,6 +290,10 @@ async Task IOrchestrationService.DeleteAsync()
...
+            // See: https://github.com/microsoft/durabletask-netherite/issues/352
+            this.startupException = null;
```
cgillum, line 295: My main concern with this approach is that we lose the benefit of caching the exception to begin with, which I assume was to prevent a non-stop storm of exceptions to occur when many concurrent clients try to start the engine at the same time. If we want to preserve the spirit of the original design, we may need a slightly more sophisticated approach where all *concurrent* calls use the cached exception, but any *subsequent* calls after the initial waves (i.e. the retries) get a fresh chance to start the backend. There are some techniques for this in C#, but it involves a small amount of wizardry.

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
davidmrdavid: Good point. I've called this out in the comments.

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
davidmrdavid: I'm not so certain (yet) that we lose the benefit here - I thought that subsequent calls to `GetClientAsync` would still fail with the expected cached error, as per:

https://github.com/microsoft/durabletask-netherite/blob/733005c5ccf7e76f815539ad6e6f2d536e058d9f/src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs#L65-L78

I was hoping that this cached error would remain _up until_ the Host calls `StartAsync` again. Am I missing something here?

In any case, I can see how some concurrency control could help us gain confidence in this implementation, so let me see if I can conjure up _this wizardry you speak of_ :-) . I can imagine some kind of concurrency-safe "generation" counter being used to make this work, unless you were thinking of something else.

sebastianburckhardt:
I just pushed some changes to this PR since I had already started working on this. However, the discussion in this PR is important so I wanted to preserve it. Basically these changes fix the potential race when resetting the state. Also, resetting the state required a bit more work than just clearing the exception.

davidmrdavid:
@sebastianburckhardt: you mentioned earlier today your implementation still had a known bug. Is that still the state of this branch?

sebastianburckhardt:
>  or is it ready for review?

It is ready for review.

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
```diff
@@ -347,20 +359,39 @@ public async Task TryStartAsync(bool clientOnly)
...
 
-        async Task<ServiceState> StartClientAsync(Task<bool> greenLight)
+        static void TryTransition<T>(ref Task<T> transition, Task<T> lastTransition, Func<Task<T>> nextTransition)
         {
-            if (!await greenLight) return ServiceState.None; 
+            var greenLight = new TaskCompletionSource<bool>();
+            var task = conditionalTransition();
+            var interlockedResult = Interlocked.CompareExchange<Task<T>>(ref transition, task, lastTransition);
+            bool success = interlockedResult == lastTransition;
+            greenLight.SetResult(success);
```
davidmrdavid, line 368: I think we should provide guiding comments here, as this is the key change from what I gather.

```suggestion

            // construct a "green light" task to control whether or not we're allowed to take the next transitoin step
            var greenLight = new TaskCompletionSource<bool>();
            var task = conditionalTransition();
            
            // If the current transition task is the same as `lastTransition`, the transition at the beginning of this operation
            // then it is safe to invoke the next transition task. If so, we set `transition` to `task`, the next transition.
            var detectedTransition = Interlocked.CompareExchange<Task<T>>(ref transition, task, lastTransition);
            bool isSafeToTransition = detectedTransition  == lastTransition;
            greenLight.SetResult(isSafeToTransition);
```

davidmrdavid (commented):
A suggestion, but the change makes sense.

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
davidmrdavid: resolving as this PR was "taken over".

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
davidmrdavid: resolving as this PR was "taken over".

src/DurableTask.Netherite/StorageLayer/Faster/FaultInjector.cs
```diff
@@ -22,6 +22,7 @@ public enum InjectionMode
         {
             None,
             IncrementSuccessRuns,
+            PermanentFail,
```
sebastianburckhardt, line 25: I think the name should indicate that this flag is injecting specifically on the client start path (unlike the other flags which apply to all calls). Perhaps `FailClientStartup` ?

src/DurableTask.Netherite/StorageLayer/Faster/FaultInjector.cs
```diff
@@ -78,6 +83,14 @@ public void Dispose()
         readonly Dictionary<int, TaskCompletionSource<object>> startupWaiters = new Dictionary<int, TaskCompletionSource<object>>();
         readonly HashSet<BlobManager> startedPartitions = new HashSet<BlobManager>();
 
+        public void StartClient()
```
sebastianburckhardt, line 86: I would use same name as for flag, so if the flag is called `FailClientStartup` lets call this `ClientStartup`

test/DurableTask.Netherite.Tests/FaultInjectionTests.cs
```diff
@@ -64,6 +64,42 @@ public async Task InjectStartup()
...
+            HostFixture fixture = null;
+
+            // first force a permnanent failure
+            using (this.faultInjector.WithMode(Faster.FaultInjector.InjectionMode.PermanentFail, injectDuringStartup: true))
+            {
+                try
+                {
+                    fixture = new HostFixture(this.settings, useCacheDebugger: true, useReplayChecker: true, restrictMemory: null, output: (msg) => this.outputHelper.WriteLine(msg));
+                    await fixture.Host.StartAsync();
+                }
```
sebastianburckhardt, line 80: the test should check here that an exception is actually thrown.

But I think the best way to do this is actually to just use the `Assert.ThrowsAsync` construct instead of a try clause:

```csharp
await Assert.ThrowsAsync<Exception>(async () =>
{
    ...
});
```

sebastianburckhardt (changes requested):

test/DurableTask.Netherite.Tests/FaultInjectionTests.cs
davidmrdavid: good point, I tend to forget this utility exists. I've refactored this accordingly

src/DurableTask.Netherite/StorageLayer/Faster/FaultInjector.cs
davidmrdavid: good idea, applied this feedback.

src/DurableTask.Netherite/StorageLayer/Faster/FaultInjector.cs
davidmrdavid: renamed accordingly

src/common.props
```diff
@@ -4,9 +4,9 @@
   <PropertyGroup>
 	<MajorVersion>1</MajorVersion>
 	<MinorVersion>5</MinorVersion>
-	<PatchVersion>3</PatchVersion>
+	<PatchVersion>4</PatchVersion>
```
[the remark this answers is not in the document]
davidmrdavid: The patch version change was intended, but not the suffix. I just reverted the whole thing though.

sebastianburckhardt (approved):

PR #377, opened 2024-04-09
Added some readmes and doc comments

cgillum:
These are some notes I took while watching the two recordings of the Netherite 101 sessions (both of which I unfortunately missed because I was on vacation). I thought it would be good to have them in the repo so that this additional context can be visible to anyone in an obvious location.

I suppose we could have put some of this content in the Netherite docs content, but I feel like this information is more suited for code maintainers rather than end users, hence it would make more sense to collocate it with the code.

@sebastianburckhardt I'm not 100% confident in the accuracy of the content I wrote, so it would be great if you could correct any mistakes. Feel free to suggest additional content as well. You can also just edit this PR directly if you want. :)

### DIFF
--- a/src/DurableTask.Netherite/Events/PartitionEvents/Internal/BatchProcessed.cs
+++ b/src/DurableTask.Netherite/Events/PartitionEvents/Internal/BatchProcessed.cs
@@ -10,6 +10,9 @@ namespace DurableTask.Netherite
     using DurableTask.Core;
     using DurableTask.Core.History;
 
+    /// <summary>
+    /// Update event that executes when a orchestration work item is completed.
+    /// </summary>
     [DataContract]
     class BatchProcessed : PartitionUpdateEvent, IRequiresPrefetch
     {
--- a/src/DurableTask.Netherite/Events/README.md
+++ b/src/DurableTask.Netherite/Events/README.md
@@ -1,0 +1,53 @@
+# Netherite Events
+
+This directory contains the event definitions for the Netherite engine. All components in Netherite communicate via events over an event bus (i.e. Azure Event Hubs). The events are defined as .NET classes that inherit from the `DurableTask.Netherite.Event` abstract base class.
+
+Event classes are organized on the file system based on where they execute. For example `ClientEvents` are events that execute on the client and `LoadMonitor` events are executed on the global load monitor component.
+
+## Event types
+
+The following are the different event types supported by Netherite.
+
+### [Client events](./ClientEvents/)
+
+Client events are events that are sent from the partition and *executed on the client*. In most cases, they are responses to client-initiated requests. For example, a client sends a [StateRequestReceived](./PartitionEvents/External/FromClients/StateRequestReceived.cs) event to a partition to request the state of an orchestration, and the partition responds back to the client with a [StateResponseReceived](./ClientEvents/StateResponseReceived.cs) event, containing that result.
+
+### [Partition events](./PartitionEvents/)
+
+*Partition events* are executed on the partition (i.e. on the worker that owns the partition). They come from clients, the load monitor, or other partitions. Partitions also send events to themselves (internal events). The partition event class definitions are organized on the file system based on where they come from.
+
+There are three types of partition events, which are used to interact with partition state:
+
+1. **Update events**: Atomically read/update one or more tracked objects. These are deterministically replay-able - i.e. it must always have the same effect.
+2. **Read events**: Reads a single object.
+3. **Query events**: Scans all `InstanceState` objects. These implement the query functionality.
+
+### [Load monitor events](./LoadMonitorEvents/)
+
+Load monitor events are sent by the partition and executed on the global load monitor component.
+
+### [Event fragments](./Fragments/)
+
+(Appears to be a special type of event that allows taking large events and breaking them up into smaller events - more research is needed to confirm)
+
+## Event processing
+
+Netherite operations that modify state are often composed of multiple events in a sequence. This is typically done because partition state needs to be loaded into the cache before it can be modified. For example, the process of creating a new orchestration or entity (instance) involves three steps:
+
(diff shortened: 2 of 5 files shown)


src/DurableTask.Netherite/Events/PartitionEvents/Internal/BatchProcessed.cs
```diff
@@ -10,6 +10,9 @@ namespace DurableTask.Netherite
     using DurableTask.Core;
     using DurableTask.Core.History;
 
+    /// <summary>
+    /// Update event that executes when a orchestration work item is completed.
+    /// </summary>
```
davidmrdavid, line 15: based on my assumptions, I presume a _batch_ refers to multiple WIs.
```suggestion
    /// <summary>
    /// Update event that executes when a orchestration work items are completed.
    /// </summary>
```

src/DurableTask.Netherite/Events/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+Event classes are organized on the file system based on where they execute. For example `ClientEvents` are events that execute on the client and `LoadMonitor` events are executed on the global load monitor component.
+
+## Event types
+
+The following are the different event types supported by Netherite.
+
+### [Client events](./ClientEvents/)
+
+Client events are events that are sent from the partition and *executed on the client*. In most cases, they are responses to client-initiated requests. For example, a client sends a [StateRequestReceived](./PartitionEvents/External/FromClients/StateRequestReceived.cs) event to a partition to request the state of an orchestration, and the partition responds back to the client with a [StateResponseReceived](./ClientEvents/StateResponseReceived.cs) event, containing that result.
```
davidmrdavid, line 13: Something confuses me here: This says that client events are **sent from the partition** and **executed on the client**, but then you explain that `StateRequestReceived` is **sent from the partition** and **executed in the partition** which seems contradictory.

Perhaps the way to make sense of this is that `StateRequestReceived` is not considered a client event? If so - I recommend we call that out for clarity's sake :-) .

src/DurableTask.Netherite/Events/README.md
```diff
@@ -0,0 +1,53 @@
...
+2. **Read events**: Reads a single object.
+3. **Query events**: Scans all `InstanceState` objects. These implement the query functionality.
+
+### [Load monitor events](./LoadMonitorEvents/)
+
+Load monitor events are sent by the partition and executed on the global load monitor component.
+
+### [Event fragments](./Fragments/)
+
+(Appears to be a special type of event that allows taking large events and breaking them up into smaller events - more research is needed to confirm)
```
davidmrdavid, line 31: If I understand this correctly, I believe you're right. If I'm not mistaken - sometimes an event is split into multiple **blobs** and as such it's events are received in fragments that need to be re-assembled.

Related: https://github.com/microsoft/durabletask-netherite/blob/main/src/DurableTask.Netherite/Util/FragmentationAndReassembly.cs

src/DurableTask.Netherite/Events/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Event processing
+
+Netherite operations that modify state are often composed of multiple events in a sequence. This is typically done because partition state needs to be loaded into the cache before it can be modified. For example, the process of creating a new orchestration or entity (instance) involves three steps:
```
davidmrdavid, line 35: I'm intrigued by the word "cache" here - is that particularly meaningful here? We can't we just say "loaded into memory"?

src/DurableTask.Netherite/Events/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+The reason for this design is that the partition state is not always in memory. It is stored in **FasterKV** and is only loaded into memory when needed. If an attempt is made to update some state when it is not in memory, the update will be queued until the state is loaded into memory, stalling the pipeline of (ordered) events the follows the update.
+
+### Partition event pipeline
+
+Events in a Netherite partition are processed in a pipeline that's composed of various "workers" once they are submitted to `IPartitionState.Submit(...)`.
+
+* `IntakeWorker`: Assigns commit log position to each event and decides whether to send the event to a `LogWorker` or a `StoreWorker`.
+* `LogWorker`: Continuously persists update events to the commit log using **FasterLog**.
+* `StoreWorker`: Handles read/update/query events on the partition state, and periodically/asynchronously checkpoints to **FasterKV**.
```
davidmrdavid, line 51: Something worth calling out as we get into the nitty gritty - FasterKV also has an inner log (called the hybrid log) and this is important to remember in particularly hairy FASTER investigations. So when we talk about "the FASTER log" it's not always immediately clear if we mean `FasterLog` of `FasterKV`'s "hybrid log". Just FYI, in case we want to sneak that info somewhere in here.

davidmrdavid (commented):
Super helpful, I feel like I learned some stuff as well - perhaps I should re-review those records as well. Left some notes, but I'll defer to Sebastian for the final call on these.

src/DurableTask.Netherite/Events/README.md
cgillum: This was the terminology that Sebastian used pretty consistently throughout the part 2 presentation so I was trying to be consistent. That said, I too don't know if there's any distinction between "cached" and "in-memory" data and prefer the latter if there isn't a useful distinction.

src/DurableTask.Netherite/Events/README.md
cgillum: It's interesting to know that FasterKV is internally implemented using a log, but I'll defer to Sebastian on whether we need additional clarity on the terminology here. I wouldn't expect the average Netherite developer to know the internals of FASTER, but then again, every Netherite contributor I know *does* know the internals of FASTER so the data already contradicts my assumptions. :)

src/DurableTask.Netherite/Events/README.md
cgillum: Yeah, I was worried that readers might be confused by this, and you confirmed it. :) Let me try to clarify this...

PR #380, opened 2024-04-15
fix race condition on OrchestrationRuntimeState.NewEvents

sebastianburckhardt:
In the current code, the `BatchProcessed` event was copying the list `NewEvents` from `OrchestrationRuntimeState`. However, that list can be mutated when processing more batches. This violates the design principle that events should be immutable, and can lead to race conditions between code that is reading those events (e.g. for serialization) and code that is mutating it.

To fix this, we just make a copy of that list when constructing the `BatchProcessed` event.

### DIFF
--- a/src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
+++ b/src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
@@ -876,7 +876,7 @@ namespace DurableTask.Netherite
                 InstanceId = workItem.InstanceId,
                 BatchStartPosition = messageBatch.BatchStartPosition,
                 BatchLength = messageBatch.BatchLength,
-                NewEvents = (List<HistoryEvent>)newOrchestrationRuntimeState.NewEvents,
+                NewEvents = newOrchestrationRuntimeState.NewEvents.ToList(),
                 WorkItemForReuse = cacheWorkItemForReuse ? orchestrationWorkItem : null,
                 PackPartitionTaskMessages = partition.Settings.PackPartitionTaskMessages,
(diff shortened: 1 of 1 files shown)


src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
```diff
@@ -876,7 +876,7 @@ async Task<PurgeResult> IOrchestrationServicePurgeClient.PurgeInstanceStateAsync
                 InstanceId = workItem.InstanceId,
                 BatchStartPosition = messageBatch.BatchStartPosition,
                 BatchLength = messageBatch.BatchLength,
-                NewEvents = (List<HistoryEvent>)newOrchestrationRuntimeState.NewEvents,
```
davidmrdavid, line 879: I wonder if we can (or should) statically enforce that `NewEvents` should be `readonly`. That way, we can have more confidence that it is immutable. Is that worth it here?

src/DurableTask.Netherite/OrchestrationService/NetheriteOrchestrationService.cs
```diff
@@ -876,7 +876,7 @@ async Task<PurgeResult> IOrchestrationServicePurgeClient.PurgeInstanceStateAsync
...
+                NewEvents = newOrchestrationRuntimeState.NewEvents.ToList(),
```
davidmrdavid, line 879: nit -> let's explain the change to avoid losing this intuition
```suggestion
                NewEvents = newOrchestrationRuntimeState.NewEvents.ToList(), // `NewEvents` in `newOrchestrationRuntimeState` may be mutated, so we copy to avoid a surprise change.
```

davidmrdavid (approved):

PR #381, opened 2024-04-16
Deduplicate CreationRequestReceived events

sebastianburckhardt:
Discovered more fallout from EH internal event duplication (as first observed in #379).

Specifically, it is possible for a creation request to be processed twice, thereby starting a duplicate work item, and this duplicate work item later creates corruption in the history (multiple execution started events) since both work items have the same execution id.

To fix this I stores the request id of the creation request that created the orchestration instance, and use that to filter duplicate deliveries of the same event.

### DIFF
--- a/src/DurableTask.Netherite/PartitionState/InstanceState.cs
+++ b/src/DurableTask.Netherite/PartitionState/InstanceState.cs
@@ -54,7 +54,7 @@ namespace DurableTask.Netherite
                 && creationRequestReceived.DedupeStatuses != null
                 && creationRequestReceived.DedupeStatuses.Contains(this.OrchestrationState.OrchestrationStatus);
 
-            if (! previousExecutionWithDedupeStatus)
+            if (!previousExecutionWithDedupeStatus)
             {
                 var ee = creationRequestReceived.ExecutionStartedEvent;
 
--- a/src/DurableTask.Netherite/PartitionState/InstanceState.cs
+++ b/src/DurableTask.Netherite/PartitionState/InstanceState.cs
@@ -27,6 +27,9 @@ namespace DurableTask.Netherite
         [DataMember]
         public List<WaitRequestReceived> Waiters { get; set; }
 
+        [DataMember]
+        public string CreationRequestEventId { get; set; }
+
         [IgnoreDataMember]
         public override TrackedObjectKey Key => new TrackedObjectKey(TrackedObjectKey.TrackedObjectType.Instance, this.InstanceId);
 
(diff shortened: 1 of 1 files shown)


src/DurableTask.Netherite/PartitionState/InstanceState.cs
```diff
@@ -39,12 +42,19 @@ public override string ToString()
...
+
             bool exists = this.OrchestrationState != null;
-            bool filterDuplicate = exists
+
+            bool previousExecutionWithDedupeStatus = exists
                 && creationRequestReceived.DedupeStatuses != null
                 && creationRequestReceived.DedupeStatuses.Contains(this.OrchestrationState.OrchestrationStatus);
 
-            if (!filterDuplicate)
+            if (! previousExecutionWithDedupeStatus)
```
davidmrdavid, line 57:
```suggestion
            if (!previousExecutionWithDedupeStatus)
```

src/DurableTask.Netherite/PartitionState/InstanceState.cs
```diff
@@ -211,6 +222,8 @@ void DeleteState(EffectTracker effects)
             effects.AddDeletion(TrackedObjectKey.History(this.InstanceId));
 
             this.OrchestrationState = null;
+            this.OrchestrationStateSize = 0;
```
davidmrdavid, line 225: was not setting this to `0` this causing an issue before?

src/DurableTask.Netherite/PartitionState/InstanceState.cs
```diff
@@ -39,12 +42,19 @@ public override string ToString()
 
         public override void Process(CreationRequestReceived creationRequestReceived, EffectTracker effects)
         {
+            if (creationRequestReceived.EventIdString == this.CreationRequestEventId)
```
davidmrdavid, line 45: I wonder if there's a way to deduplicate EH earlier in the pipeline, so that we don't have to sprinkle de-duplication logic in various places. Not sure if there's some kind of counter we can use to determine that we've processed all events up to some ID, or something like that. Hard to say, I could also see the possibility that we need to handle each case separately.

davidmrdavid (approved):

davidmrdavid:
@sebastianburckhardt: I'd love to get this merged for the upcoming release, and I think we can mostly ignore my nitpicks. If this looks good to merge, please go ahead!

src/DurableTask.Netherite/PartitionState/InstanceState.cs
sebastianburckhardt: I don't think there is anything we can do here, because we cannot keep track of the lifetime of clients.

PR #385, opened 2024-04-19
Update to latest event hubs client SDK

sebastianburckhardt:
Update to the latest Event Hubs SDK, to remove our dependency on deprecated packages.

This was not straightforward: lots of API surface changes, and some minor functionality changes as a result. In the end, the solution seems reasonably similar though and better in some ways. 

1. I was able to remove the EH-internal checkpointing entirely (which was inefficient and superfluous) since EH now allows me to tell them what position to resume at when a partition is opened.
2. I found a way to still allow batched processing of events by overriding a virtual function in the client.

For motivation and discussion, see https://github.com/microsoft/durabletask-netherite/issues/310.

Fixes https://github.com/microsoft/durabletask-netherite/issues/310.

### DIFF
--- a/src/DurableTask.Netherite/DurableTask.Netherite.csproj
+++ b/src/DurableTask.Netherite/DurableTask.Netherite.csproj
@@ -25,8 +25,8 @@
   <!-- Version settings: https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/ -->
   <PropertyGroup>
     <MajorVersion>1</MajorVersion>
-	<MinorVersion>5</MinorVersion>
-	<PatchVersion>0</PatchVersion>
+	<MinorVersion>4</MinorVersion>
+	<PatchVersion>3</PatchVersion>
 	<VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
     <VersionSuffix></VersionSuffix>
     <AssemblyVersion>$(MajorVersion).0.0.0</AssemblyVersion>
--- a/src/DurableTask.Netherite/DurableTask.Netherite.csproj
+++ b/src/DurableTask.Netherite/DurableTask.Netherite.csproj
@@ -42,8 +42,8 @@
   <ItemGroup>
     <PackageReference Include="Azure.Core" Version="1.38.0" />
     <PackageReference Include="Azure.Data.Tables" Version="12.8.0" />
-		<PackageReference Include="Azure.Messaging.EventHubs" Version="5.11.2" />
-		<PackageReference Include="Azure.Messaging.EventHubs.Processor" Version="5.11.2" />
+    <PackageReference Include="Azure.Messaging.EventHubs" Version="5.11.2" />
+    <PackageReference Include="Azure.Messaging.EventHubs.Processor" Version="5.11.2" />
     <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="11.2.3" />
     <PackageReference Include="Azure.Storage.Blobs" Version="12.16.0" />
     <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.16.2" />
--- a/src/DurableTask.Netherite/DurableTask.Netherite.csproj
+++ b/src/DurableTask.Netherite/DurableTask.Netherite.csproj
@@ -50,12 +50,12 @@
   </ItemGroup>
   
   <ItemGroup>
-    <PackageReference Include="Azure.Core" Version="1.33.0" />
+    <PackageReference Include="Azure.Core" Version="1.38.0" />
     <PackageReference Include="Azure.Data.Tables" Version="12.8.0" />
-    <PackageReference Include="Azure.Messaging.EventHubs" Version="5.9.2" />
-    <PackageReference Include="Microsoft.Azure.EventHubs.Processor" Version="4.3.2" />
-    <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="11.2.3" />
-    <PackageReference Include="Azure.Storage.Blobs" Version="12.16.0" />
+    <PackageReference Include="Azure.Messaging.EventHubs" Version="5.11.2" />
+    <PackageReference Include="Azure.Messaging.EventHubs.Processor" Version="5.11.2" />
+		<PackageReference Include="Microsoft.Azure.Storage.Blob" Version="11.2.3" />
+		<PackageReference Include="Azure.Storage.Blobs" Version="12.16.0" />
 	<PackageReference Include="Microsoft.FASTER.Core" Version="2.0.23" />
     <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.15.1" />
 	<PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
--- a/src/DurableTask.Netherite/DurableTask.Netherite.csproj
+++ b/src/DurableTask.Netherite/DurableTask.Netherite.csproj
@@ -20,8 +20,17 @@
     <PackageId>Microsoft.Azure.DurableTask.Netherite</PackageId>
     <PackageIcon>icon.png</PackageIcon>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+	<PackageReadmeFile>README.md</PackageReadmeFile>
   </PropertyGroup>
 
+  <!-- Include the README.md at the root of the repo into the nupkg-->
+  <ItemGroup>
+    <None Include="..\..\README.md">
+      <Pack>True</Pack>
+      <PackagePath>\</PackagePath>
+    </None>
+  </ItemGroup>
+	
   <!-- Version can be edited in common.props -->
   <Import Project=".\..\common.props" />
   <ItemGroup>
--- a/src/DurableTask.Netherite/DurableTask.Netherite.csproj
+++ b/src/DurableTask.Netherite/DurableTask.Netherite.csproj
@@ -22,18 +22,8 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
-  <!-- Version settings: https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/ -->
-  <PropertyGroup>
-    <MajorVersion>1</MajorVersion>
-	<MinorVersion>4</MinorVersion>
-	<PatchVersion>3</PatchVersion>
-	<VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
-    <VersionSuffix></VersionSuffix>
-    <AssemblyVersion>$(MajorVersion).0.0.0</AssemblyVersion>
-    <BuildSuffix Condition="'$(GITHUB_RUN_NUMBER)' != ''">.$(GITHUB_RUN_NUMBER)</BuildSuffix>
-    <FileVersion>$(VersionPrefix)$(BuildSuffix)</FileVersion>
-  </PropertyGroup>
-  
+  <!-- Version can be edited in common.props -->
+  <Import Project=".\..\common.props" />
   <ItemGroup>
     <None Include="icon.png" Pack="true" PackagePath="\" />
   </ItemGroup>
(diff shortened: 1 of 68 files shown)


akhleshg:
This change has been in a pending state for a long time. Is there any target date for this fix to get checked-In? 

I am asking this because our service started using Netherite but If we don't have timeline for fix, then to meet compliance requirement, we have to move aways from using Netherite package.

akhleshg:
This change has been in a pending state for a long time. Is there any target date for this fix to get checked-In? 

I am asking this because our service started using Netherite but If we don't have timeline for fix, then to meet compliance requirement, we have to move aways from using Netherite package.

saguiitay:
@sebastianburckhardt any ETA for this PR?

akhleshg:
@sebastianburckhardt, do you have estimate timeline to chekiin this PR?

sebastianburckhardt:
Tests seem to be running o.k.

davidmrdavid:
@akhleshg, @saguiitay: apologies for the delayed response here. This is definitely in our radar, but it's release is scheduled to occur after some breaking changes we're releasing in lower-level dependencies we often, namely DurableTask.Core (we're releasing v3) and the DurableTask Extension (also releasing v3).

DurableTask.Core v3 released yesterday: https://github.com/Azure/durabletask/releases/tag/durabletask.core-v3.0.0
Durable Extension v3 is scheduled to release in late August, due to some Azure infrastructure updates we depend on for auto-scaling.

At that point, Netherite will upgrade these dependencies to v3, and the Event Hubs dependency, and increase it's own major version to reflect these breaking changes.

That said - @akhleshg, @saguiitay : are you all using Durable Functions Netherite or DurableTask Netherite. If you're using DurableTask directly, we may be able to expedite this by a few weeks as that would only require upgrading **DurableTask.Core**, which is already out.

akhleshg:
Thank David (@davidmrdavid) for your response and for providing additional details about the changes. We are currently using DurableTask Netherite and require the update to ensure security compliance. I would greatly appreciate it if you could expedite the necessary changes to DurableTask Netherite, as we are eagerly awaiting this update.


We do have Durable Functions and plan to update them to use Netherite. However, we can wait for the Durable Functions Netherite update before migrating our service to Netherite.

anttix:
> If you're using DurableTask directly, we may be able to expedite this by a few weeks as that would only require upgrading **DurableTask.Core**, which is already out.

Hi @davidmrdavid ! We are in the early stages of adoption of DurableTask Netherite and ran into #417 so we could make use of that expedited release if possible :) Thank you!

davidmrdavid:
Looking to unblock this, trying to see what we can do

src/DurableTask.Netherite/TransportLayer/EventHubs/PartitionProcessor.cs
```diff
@@ -92,70 +95,81 @@ public EventHubsProcessor(
...
+                        this.traceHelper.LogTrace("EventHubsProcessor {eventHubName}/{eventHubPartition} discarding buffered event nextInputQueuePosition={nextInputQueuePosition} lastInBatch={lastInBatch} seqno={seqNo} batchBlob={batchBlob}", this.eventHubName, this.eventHubPartition, front.Event.NextInputQueuePositionTuple, confirmed.LastInBatch, confirmed.SeqNo, confirmed.BatchBlob?.Name);
                     }
 
                     if (confirmed.LastInBatch)
                     {
-                        // to create EH checkpoints we need to record both the EH sequence number and the EH offset (see SaveEventHubsReceiverCheckpoint)
-                        this.persistedOffset = Math.Max(this.persistedOffset, confirmed.Offset);
-                        this.persistedSequenceNumber = Math.Max(this.persistedSequenceNumber, confirmed.SeqNo);
-
```
davidmrdavid, line 158: why isn't this needed anymore?

src/DurableTask.Netherite/TransportLayer/EventHubs/PartitionProcessor.cs
```diff
@@ -255,30 +269,36 @@ async Task<PartitionIncarnation> StartPartitionAsync(PartitionIncarnation prior
...
+                            c.Partition.SubmitEvents(batch);
+                            this.traceHelper.LogDebug("EventHubsProcessor {eventHubName}/{eventHubPartition} redelivered {batchsize} packets, starting with #{seqno}, next expected packet is #{nextSeqno} (incarnation {incarnation})", this.eventHubName, this.eventHubPartition, batch.Count, batch[0].NextInputQueuePosition - 1, c.NextPacketToReceive, c.Incarnation);
+                        }
+                        else
+                        {
+                            this.traceHelper.LogDebug("EventHubsProcessor {eventHubName}/{eventHubPartition} has no packets requiring redelivery ", this.eventHubName, this.eventHubPartition);
+                        }
                     }
-                }
+                }
```
davidmrdavid, line 297:
```suggestion
                } 
```

src/DurableTask.Netherite/TransportLayer/EventHubs/PartitionProcessor.cs
```diff
@@ -293,142 +313,133 @@ async Task<PartitionIncarnation> StartPartitionAsync(PartitionIncarnation prior
...
-                else
-                {
-                    this.traceHelper.LogDebug("EventHubsProcessor {eventHubName}/{eventHubPartition} shutdown task already started", this.eventHubName, this.eventHubPartition);
+                    this.shutdownRegistration?.Dispose();
+
+                    // dispose the lock, after quickly acquiring and releasing it to make sure no one is still holding it
+                    (await this.deliveryLock.LockAsync()).Dispose();
+                    this.deliveryLock.Dispose();
+
+                    this.traceHelper.LogInformation("EventHubsProcessor {eventHubName}/{eventHubPartition} is shut down", this.eventHubName, this.eventHubPartition);
```
davidmrdavid, line 361: this is a bit strange. Can you please clarify, in the code, why this is needed?

davidmrdavid: Or ideally, can the code be refactored so we don't need this trick? I understand if not, it's just a bit odd

davidmrdavid (commented):
Left some comments. The `PartitionProcessor` is a bit more complex than I would like (thus hard to review), but I understand this is not the right time to ask for a bit refactoring there. But it would be good to review it as a team to see if we can either simplify it or at least help document it.

I think we can be practical and consider the addition of the new SDK as not a breaking change, even though in the strictest sense it is. We'll see, let's discuss internally.

One test we should absolutely do is ensure old generated by the old EH SDK is compatible with the new EH SDK data, and vice-versa. This was a problem in the Azure Storage SDK upgrade we're doing in DTFx and the WebJobs extension.

src/DurableTask.Netherite/DurableTask.Netherite.csproj
```diff
@@ -40,10 +40,10 @@
   </ItemGroup>
   
   <ItemGroup>
-    <PackageReference Include="Azure.Core" Version="1.33.0" />
+    <PackageReference Include="Azure.Core" Version="1.38.0" />
     <PackageReference Include="Azure.Data.Tables" Version="12.8.0" />
-    <PackageReference Include="Azure.Messaging.EventHubs" Version="5.9.2" />
-    <PackageReference Include="Microsoft.Azure.EventHubs.Processor" Version="4.3.2" />
+		<PackageReference Include="Azure.Messaging.EventHubs" Version="5.11.2" />
+		<PackageReference Include="Azure.Messaging.EventHubs.Processor" Version="5.11.2" />
```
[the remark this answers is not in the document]
sebastianburckhardt: Looks right on my screen (for me in VS tabs appear as 2 spaces in csproj files) but not for someone where tabs are displayed as 4 spaces.

Ideally I would want VS to replace all tabs with spaces (it does this in C# already) but I suppose it would need some kind of configuration.

src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsTransport.cs
```diff
@@ -214,8 +210,11 @@ internal async Task ExitProcess(bool deletePartitionsFirst)
                 await this.connections.DeletePartitions();
             }
 
-            this.traceHelper.LogError("EventHubsTransport is killing process in 10 seconds");
-            await Task.Delay(TimeSpan.FromSeconds(10));
+            this.traceHelper.LogError("EventHubsTransport is killing process in 5 seconds");
+            
+            Task _ = ((ITransportLayer)this).StopAsync(true);
+
+            await Task.Delay(TimeSpan.FromSeconds(5));
```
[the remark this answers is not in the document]
sebastianburckhardt: I think it is so the leases get released before exiting the process. This speeds up subsequent recovery.

src/DurableTask.Netherite/TransportLayer/EventHubs/PartitionProcessor.cs
sebastianburckhardt: It looks slightly wrong (double Dispose). Will fix that.
Also I will add a comment explaining it a bit more.

src/DurableTask.Netherite/TransportLayer/EventHubs/PartitionProcessor.cs
sebastianburckhardt: with the new EH SDK, we are no longer forced to use the EH-built-in-checkpointing. Instead we can, at partition startup time. specify the position at which we would start to receive events from. 

This is much simpler for us since we already have that information stored in our partition state.  So we can avoid those unnecessary extra checkpoints that EH forced us to do.

sebastianburckhardt:
*
The PartitionProcessor is a bit more complex than I would like (thus hard to review), but I understand this is not the right time to ask for a bit refactoring there. But it would be good to review it as a team to see if we can either simplify it or at least help document it.

The main source of complexity is the difficulty of reconciling the API that EH gives us with how Netherite manages partitions.

In particular the 'reincarnation' mechanism is complicated. The main problem here is that we cannot tell EH to terminate a partition. So we have to present this "never failing partition" surface to the EH API but internally partitions can fail, thus the reincarnation mechanism.

I don't think there are any easy solutions to this problem and we have spent a lot of time debugging this current implementation already. If we replace it with something new, we may just run into a lot of new bugs to fix which may not be worth it.

One thing at least got simpler (we don't need to use the EH checkpoints anymore).

Sebastian

________________________________

src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsTransport.cs
```diff
@@ -214,8 +210,11 @@ internal async Task ExitProcess(bool deletePartitionsFirst)
                 await this.connections.DeletePartitions();
             }
 
-            this.traceHelper.LogError("EventHubsTransport is killing process in 10 seconds");
-            await Task.Delay(TimeSpan.FromSeconds(10));
+            this.traceHelper.LogError("EventHubsTransport is killing process in 5 seconds");
+            
+            Task _ = ((ITransportLayer)this).StopAsync(true);
```
davidmrdavid, line 215:
```suggestion
            // release leases before exiting the process. This speeds up subsequent recovery.
            Task _ = ((ITransportLayer)this).StopAsync(true);
```

src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsTransport.cs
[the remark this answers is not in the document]
davidmrdavid: Understood. In that case, let me replace this comment with the following suggestion: https://github.com/microsoft/durabletask-netherite/pull/385/files#r1772166687

davidmrdavid:
Almost done, just need another pass over the PartitionProcessor before approving, all other files have been reviewed

davidmrdavid (approved):
LGTM. Before **releasing**, let's please rev the major version of Netherite in all csprojs.
I do want to emphasize again that I think we refactoring our PartitionProcesor would be good, it's a tricky piece of code to review. FYI @AnatoliB, I think this file would be a good place to get the eng side to contribute to and help document + refactor to make it easier to follow. I think it would pay dividends.

Anyways, I've gone over all files, looks good to me.
Before releasing, let's also make sure there's no compatibility issues with the previous EH SDK, to avoid issues like what we had with the Azure Storage SDK upgrade. We can help with this.

Issue #390, opened 2024-05-02
System.NotSupportedException: Durable entities are not supported by the current backend configuration


jaliyaudagedara:
[durabletask-netherite: 1.5.1 says it added support for `isolated entities`](https://github.com/microsoft/durabletask-netherite/releases/tag/1.5.1), but it still doesn't seem to work.

**Exception:**
```
System.NotSupportedException: Durable entities are not supported by the current backend configuration
```
Example Project: 
[AzureFunctionsDemo.Netherite.zip](https://github.com/microsoft/durabletask-netherite/files/15185294/AzureFunctionsDemo.Netherite.zip)

**csproj**
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <AzureFunctionsVersion>v4</AzureFunctionsVersion>
    <OutputType>Exe</OutputType>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>
  <ItemGroup>
    <FrameworkReference Include="Microsoft.AspNetCore.App" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.22.0" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.DurableTask" Version="1.1.2" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.DurableTask.Netherite" Version="1.5.1" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.1.0" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore" Version="1.2.1" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.17.2" />
    <PackageReference Include="Microsoft.ApplicationInsights.WorkerService" Version="2.22.0" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.ApplicationInsights" Version="1.2.0" />
  </ItemGroup>
  <ItemGroup>
    <None Update="host.json">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
    </None>
    <None Update="local.settings.json">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
[...]

PR #393, opened 2024-05-03
Auto-generate AssemblyInfo + refactor version into common props.

davidmrdavid:
The Netherite version of this PR: https://github.com/microsoft/durabletask-mssql/pull/213

Our most recent Netherite release accidentaly missed updating the AssemblyInfo.cs file, leading to the following issue: https://github.com/microsoft/durabletask-netherite/issues/390

This PR does two things in response:
(1) Auto-generates the AssemblyInfo file, as in https://github.com/microsoft/durabletask-mssql/pull/213
(2) Refactor the version properties in a `common.props` file that is referenced by all `.csproj`s.

**Open question:** On VS, I struggled to add the common.props video to the Solution view, which would make editing it inside the IDE difficult. I'd like to fix that before merging this.


cgillum:
`common.props` works, but even simpler would be to use `Directory.Build.props`, because all project files in the same directory hierarchy get the definitions get imported automatically (you don't need to do an explicit import in each .csproj file). More info here: https://learn.microsoft.com/visualstudio/msbuild/customize-by-directory?view=vs-2022. This is the approach I've been using for all new projects. Here's a [recent example](https://github.com/Azure/azure-functions-openai-extension/blob/fa2073cb8d9c178d0a4e0248540f3cdfaa49af44/src/Directory.Build.props#L25-L28).

davidmrdavid:
Hmm, I know Sebastian had some reservations about the "magicness" of directory.build.pros, from here: https://github.com/microsoft/durabletask-netherite/pull/224#issuecomment-1424971053 . So although I'd prefer to "stick to convention" and use "directory.build.props", I think using the more manual approach of this PR may be less controversial (and obvious in how it works) middle ground.

src/common.props
jviau, line 1: You can create a file `Directory.Build.props` and it is auto-imported by all projects in its cone. 

https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-by-directory?view=vs-2022

jviau:
> Hmm, I know Sebastian had some reservations about the "magic" of directory.build.props, from here: [#224 (comment)](https://github.com/microsoft/durabletask-netherite/pull/224#issuecomment-1424971053) . So although I'd prefer to "stick to convention" and use "directory.build.props", I think using the more manual approach of this PR may be less controversial (and obvious in how it works).

But there is also _so_ much more to msbuild 'magic' than Directory.Build.props - so why single that out? You can view the stitched together import graph by running `dotnet build -pp:pp.xml` , which will output `pp.xml`. It is thousands (tens even?) of lines long. Directory.Build.props is just a drop in the bucket. Plus, once you know how it works, is it magic anymore?

src/common.props
```diff
@@ -0,0 +1,14 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- Version settings: https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/ -->
+  <PropertyGroup>
+	<MajorVersion>1</MajorVersion>
+	<MinorVersion>5</MinorVersion>
+	<PatchVersion>1</PatchVersion>
+	<VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
+	<VersionSuffix></VersionSuffix>
+	<AssemblyVersion>$(MajorVersion).0.0.0</AssemblyVersion>
```
jviau, line 10: What does pinning assembly version to only major offer?

davidmrdavid: This a convention I think @cgillum first introduced to the DF assemblies. I recall it being related to hotswapping dlls. Can't say, I'll let him explain.

Note - I did not add this pinning to the repo, this already existed

davidmrdavid:
> Plus, once you know how it works, is it magic anymore?

I personally have no strong feelings on `Build.directory.props` vs `common.props`, I just think that since there's still an unresolved conversation about it in another thread, I shouldn't bypass that discussion in this PR. So I'm opting for a different, more "low tech" approach. Alternatively, I can just remove this part of the PR, as my main goal was to automate the AssemblyInfo metadata.

No strong feelings here. Happy to remove this part of the PR if we'd rather "do things right" from the beginning instead of taking this partial step.

src/common.props
cgillum: @jviau The original Azure Functions engineering manager who was previously on the ASP.NET team strongly recommended this to avoid problems related to compatibility issues with GAC'd assemblies. I can't remember exactly what all the issues were, but it made sense at the time and been trying to follow it.

cgillum (approved):
Given Sebastian's strong feelings about Build.Directory.props, I'm inclined to say let's just go with what you have (common.props). It's largely a matter of preference so I'd like to respect the wishes of the primary maintainer. :)

src/common.props
jviau: Ah, back in ASP.NET days yeah the GAC was relevant. I don't think it is anymore. This is safe to remove nowadays. I don't see this pattern present anymore - not in func host, or dotnet builds itself.

davidmrdavid:
FYI @sebastianburckhardt of this change. Hoping it will minimize release mishaps.

PR #395, opened 2024-05-07
Increase FASTER.Core to 2.6.5

davidmrdavid:
As in title - FASTER release made a hotfix for us here: https://github.com/microsoft/FASTER/pull/916


davidmrdavid:
Unfortunately, the CI is broken - so we'll need to resolve that first.

nytian (approved):

Issue #383, opened 2024-04-18
Large amounts of warnings from FASTER.KV


ericleigh007:
We have transitioned to Netherite in our production program and for the most part things are going well, although we don't have anywhere near as much experience as with the standard backend.

We have found in some cases, we experience a good number of errors and warnings.   The following seem to be generated from FASTER, not Netherite itself.

```
Part00 query 52eaf82R1809 attempt 2024-04-18T09:25:04.4847863Z enumeration failed with exception System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at FASTER.core.FasterKV`2.HandleOperationStatus[Input,Output,Context](FasterExecutionContext`3 opCtx, PendingContext`3& pendingContext, OperationStatus operationStatus, AsyncIOContext`2& request)
   at DurableTask.Netherite.Faster.FasterKV.ReadOnQuerySessionAsync(String instanceId, CancellationToken cancellationToken) in /_/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs:line 59
   at DurableTask.Netherite.Faster.FasterKV.QueryEnumeratedStates(EffectTracker effectTracker, PartitionQueryEvent queryEvent, IEnumerator`1 enumerator, Int32 pageSize, TimeSpan timeBudget, DateTime attempt)+MoveNext() in /_/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs:line 1047

Part05.35545933428    query 52eaf82R1796 attempt 2024-04-18T09:25:00.1713018Z enumeration failed with exception System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at FASTER.core.FasterKV`2.HandleOperationStatus[Input,Output,Context](FasterExecutionContext`3 opCtx, PendingContext`3& pendingContext, OperationStatus operationStatus, AsyncIOContext`2& request)
   at DurableTask.Netherite.Faster.FasterKV.ReadOnQuerySessionAsync(String instanceId, CancellationToken cancellationToken) in /_/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs:line 59
[...]

PR #397, opened 2024-05-08
Complete FASTER sessions before releasing them during query processing

davidmrdavid:
Should fix: https://github.com/microsoft/durabletask-netherite/issues/383

### Context

Some customers are reporting large numbers of FASTER warnings of the following form:

> **Part00 query 52eaf82R1809 attempt 2024-04-18T09:25:04.4847863Z enumeration failed with exception System.IndexOutOfRangeException: Index was outside the bounds of the array.**
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at FASTER.core.FasterKV`2.HandleOperationStatus[Input,Output,Context](FasterExecutionContext`3 opCtx, PendingContext`3& pendingContext, OperationStatus operationStatus, AsyncIOContext`2& request)
   at DurableTask.Netherite.Faster.FasterKV.ReadOnQuerySessionAsync(String instanceId, CancellationToken cancellationToken) in /_/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs:line 59
   at DurableTask.Netherite.Faster.FasterKV.QueryEnumeratedStates(EffectTracker effectTracker, PartitionQueryEvent queryEvent, IEnumerator`1 enumerator, Int32 pageSize, TimeSpan timeBudget, DateTime attempt)+MoveNext() in /_/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs:line 1047

The stack trace here suggests this is occurring when processing DF client queries (like getting the status of an instanceId). 

@TedHartMS suspects this error is occurring because [two threads may be using the same FASTER session at the same time](https://github.com/microsoft/durabletask-netherite/issues/383#issuecomment-2065381453).

_We originally thought this could not be the case_ as we have tight concurrency control over FASTER sessions in `ReadOnQuerySessionAsync` (notice the use of `ConcurrentBag` to obtain sessionIDs in a thread-safe way, and the semaphore):

https://github.com/microsoft/durabletask-netherite/blob/f8b5634746585c92a59fa0019c39930925c28aa9/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs#L50-L71

However, Ted called out that we're not calling `.Complete()` on that session's result before exiting that method, which means that the FASTER session may be obtained through the `ConcurrentBag` before FASTER is done cleaning up it's previous execution data, leading to the exception reported above. Today, this `.Complete()` method is called later on, in the caller of the `ReadOnQuerySessionAsync`, here: https://github.com/microsoft/durabletask-netherite/blob/f8b5634746585c92a59fa0019c39930925c28aa9/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs#L1055-L1056

### This PR

This PR attempts to call refactor the code in `ReadOnQuerySessionAsync` and it's caller so that `.Complete` is called in `ReadOnQuerySessionAsync`  itself, hopefully preventing the exception.

It also renames some variables in relevant methods to make them, in my opinion, easier to read. I also added several comments to help future maintainers understand the logic. This is a subjective change, and I'm open to feedback.

### Request for reviewiers
I worry that there may be a reason why we weren't calling `.Complete()` in `ReadOnQuerySessionAsync`: most likely due to performance reasons? I'd appreciate a focus on possible performance regressions.

I also notice the caller of `ReadOnQuerySessionAsync` follows a "producer-consumer" pattern, where the producer issues FASTER read requests (through `ReadOnQuerySessionAsync`) and the consumer calls `.Complete()` on them. Now, the producer calls `.Complete()` and I worry I may have broken some underlying assumptions of that design. Please let me know if that seems true to you all.

### DIFF
--- a/src/DurableTask.Netherite.AzureFunctions/DurableTask.Netherite.AzureFunctions.csproj
+++ b/src/DurableTask.Netherite.AzureFunctions/DurableTask.Netherite.AzureFunctions.csproj
@@ -26,9 +26,9 @@
   <PropertyGroup>
 	<MajorVersion>1</MajorVersion>
 	<MinorVersion>5</MinorVersion>
-	<PatchVersion>1</PatchVersion>
+	<PatchVersion>2</PatchVersion>
 	<VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
-    <VersionSuffix></VersionSuffix>
+    <VersionSuffix>clientQueryFix.1</VersionSuffix>
     <AssemblyVersion>$(MajorVersion).0.0.0</AssemblyVersion>
     <BuildSuffix Condition="'$(GITHUB_RUN_NUMBER)' != ''">.$(GITHUB_RUN_NUMBER)</BuildSuffix>
     <FileVersion>$(VersionPrefix)$(BuildSuffix)</FileVersion>
--- a/src/DurableTask.Netherite.AzureFunctions/DurableTask.Netherite.AzureFunctions.csproj
+++ b/src/DurableTask.Netherite.AzureFunctions/DurableTask.Netherite.AzureFunctions.csproj
@@ -22,17 +22,9 @@
     <PackageIcon>icon.png</PackageIcon>
   </PropertyGroup>
 
-  <!-- Version settings: https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/ -->
-  <PropertyGroup>
-	<MajorVersion>1</MajorVersion>
-	<MinorVersion>5</MinorVersion>
-	<PatchVersion>2</PatchVersion>
-	<VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
-    <VersionSuffix>clientQueryFix.1</VersionSuffix>
-    <AssemblyVersion>$(MajorVersion).0.0.0</AssemblyVersion>
-    <BuildSuffix Condition="'$(GITHUB_RUN_NUMBER)' != ''">.$(GITHUB_RUN_NUMBER)</BuildSuffix>
-    <FileVersion>$(VersionPrefix)$(BuildSuffix)</FileVersion>
-  </PropertyGroup>
+  <!-- Version can be edited in common.props -->
+  <Import Project=".\..\common.props" />
+
 
   <!-- Our netcoreapp2.2 target is a non-functional dummy target, so we don't need the warning -->
   <PropertyGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.2' ">
--- a/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
+++ b/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
@@ -21,6 +21,7 @@ namespace DurableTask.Netherite.Faster
     using FASTER.core;
     using Microsoft.Azure.Storage.Blob.Protocol;
     using Newtonsoft.Json;
+    using static DurableTask.Netherite.SessionsState;
 
     class FasterKV : TrackedObjectStore
     {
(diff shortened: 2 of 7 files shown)


src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
```diff
@@ -950,45 +953,56 @@ public override ValueTask RemoveKeys(IEnumerable<TrackedObjectKey> keys)
...
 
-            var channel = Channel.CreateBounded<(bool last, ValueTask<FasterKV<Key, Value>.ReadAsyncResult<EffectTracker, Output, object>> responseTask)>(200);
-            using var leftToFill = new SemaphoreSlim(pageLimit.HasValue ? pageLimit.Value : 100);
             using var cancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(this.partition.ErrorHandler.Token);
             var cancellationToken = cancellationTokenSource.Token;
 
-            Task readIssueLoop = Task.Run(ReadIssueLoop);
-            async Task ReadIssueLoop()
+            this.partition.EventDetailTracer?.TraceEventProcessingDetail($"query {queryId} attempt {attempt:o} enumeration from {queryEvent.ContinuationToken} with pageLimit={(pageLimit.HasValue ? pageLimit.ToString() : "none")} timeBudget={timeBudget}");
+            Stopwatch stopwatch = Stopwatch.StartNew();
```
davidmrdavid, line 972: here I'm just re-ordering some variables, to help with readability (it's a long method, I'm trying to group related things together).

jviau (approved):
This overall looks safe to me. You have moved the `Complete` call from just after channel read to immediately before channel write. There is no other code (of this repo) between those two points. But more importantly, you complete the FASTER session before the finally block runs, returning the session for another to take.

I can see how it was a race condition before

davidmrdavid:
I'm going temporarily add a suffix to this PR's version to try and create a private release

src/DurableTask.Netherite.AzureFunctions/DurableTask.Netherite.AzureFunctions.csproj
```diff
@@ -26,9 +26,9 @@
   <PropertyGroup>
 	<MajorVersion>1</MajorVersion>
 	<MinorVersion>5</MinorVersion>
-	<PatchVersion>1</PatchVersion>
+	<PatchVersion>2</PatchVersion>
 	<VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
-    <VersionSuffix></VersionSuffix>
+    <VersionSuffix>clientQueryFix.1</VersionSuffix>
```
jviau, line 31: I recommend just using `preview`.

sebastianburckhardt:
> I worry that there may be a reason why we weren't calling .Complete() in ReadOnQuerySessionAsync: 

I can't think of a good reason, perhaps performance was a concern, yes, but I don't think it really matters since from what I understand calling `Complete()` should always finish quickly.

> I worry I may have broken some underlying assumptions of that design

looks fine to me.

davidmrdavid:
Wonderful. If the diff looks good (it ended up looked weirder than anticipated after I indented everything in a `try-finally` block), then I'd love to run the perf tests with this change to see if we see any of the reported warnings still (either the dictionary conflict, or the `objectDisposedException` ). If we don't, then we're good. May need assistance with that, I don't have the scripts.

davidmrdavid:
@sebastianburckhardt: can I get a review from you here? I'd like to get this in, if it looks good, on the upcoming release. We're hoping to release sometime next week

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
```diff
@@ -950,187 +953,205 @@ public override ValueTask RemoveKeys(IEnumerable<TrackedObjectKey> keys)
...
+            pageDone:
+                yield return (position, null);
+                ReportProgress("completed-page");
+            done:
+                yield break;
+            }
+            finally
+            {
+                cancellationTokenSource.Cancel();
+                await produceFasterReadsTask;
```
sebastianburckhardt, line 1151: For consistency this await should use `ConfigureAwait(false)` also, I think.

And perhaps we could add a `ReportProgress` after this await? So we can detect cases where the await hangs or where the finally block does not get called (finally block in method with `yield` gets called only if caller correctly disposes enumerator).

sebastianburckhardt (approved):

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
```diff
@@ -950,187 +953,205 @@ public override ValueTask RemoveKeys(IEnumerable<TrackedObjectKey> keys)
...
-                        leftToFill.Release();
-                    }
                 }
+            pageDone:
+                yield return (position, null);
+                ReportProgress("completed-page");
+            done:
+                yield break;
+            }
+            finally
```
sebastianburckhardt, line 1148: Is there a specific reason you added a try/finally wrapper? I suppose it is for handling additional places where exceptions could be thrown that are not already caught?

davidmrdavid: That's precisely right. I just want to make sure there's no usages of the semaphore after it's been disposed and, to do that, I need to make sure the producer task is awaited before we exit this method (as after we exit, the semaphore will be disposed)

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
davidmrdavid: Added the ConfigureAwait(false)

davidmrdavid: And added the ReportProgress on my last commit

sebastianburckhardt (approved):

PR #408, opened 2024-07-25
Use lookup algorithm for compaction 

sebastianburckhardt:
Since we moved to a newer version of FASTER we can now use a more efficient version of the compaction algorithm.

In our experiments this showed dramatic reductions in time, memory, and I/O consumption.


davidmrdavid (approved):
Assuming CI goes green, what a neat, tiny, PR

PR #409, opened 2024-07-25
Implement defensive measure against compaction loops

sebastianburckhardt:
This PR addresses a problem we observed in some of the long-running tests.

The problem is that the minimal log size (as estimated based on the number of instances) seems to be different than the actual minimal log size (as a result of full compaction without any other updates). This means that compaction is continuously triggered.

We don't as of yet have an explanation, but as a defensive measure, this PR observes the discrepancy (i.e. a compaction being triggered that has zero effect) and prevents the infinite loop.

### DIFF
--- a/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
+++ b/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
@@ -48,6 +48,8 @@ namespace DurableTask.Netherite.Faster
         static readonly SemaphoreSlim availableQuerySessions = new SemaphoreSlim(queryParallelism);
         readonly ConcurrentBag<int> idleQuerySessions = new ConcurrentBag<int>(Enumerable.Range(0, queryParallelism));
 
+        long observedMinimalLogSize;
+
         async ValueTask<(Status, Output)> ReadWithFasterAsync(string instanceId, CancellationToken cancellationToken)
         {
             await availableQuerySessions.WaitAsync();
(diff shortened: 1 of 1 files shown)


src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
```diff
@@ -599,6 +608,18 @@ void RunCompaction()
                                 this.Log.BeginAddress - beginAddressBeforeCompaction,
                                 this.GetElapsedCompactionMilliseconds());
 
+
+                            long mutableSizeAfterCompaction = this.Log.TailAddress - this.Log.SafeReadOnlyAddress;
+
+                            if (this.Log.SafeReadOnlyAddress == compactedUntil && mutableSizeBeforeCompaction == 0 && readOnlySizeBeforeCompaction == mutableSizeAfterCompaction)
```
davidmrdavid, line 614: I think this is a good opportunity to explicit document the relationship between the read-only and mutable segments with compaction. Can we write a few comments about how we expect them to change with compaction (plus a refresher on how data enters and leaves the read-only segment). I think that's important to understand why this detection works.

sebastianburckhardt: I have added extensive comments to explain compaction.

I also made turned this from a simple trace statement into a warning. This will help us assess the priority of how often this happens and serve as a reminder.

davidmrdavid (approved):
Thank you for the extremely detailed comments

PR #419, opened 2024-09-20
Add timeouts to test and fix race condition in device startup

sebastianburckhardt:
Added some timeouts to the tests since we are seeing issues with hangs in the CI pipeline.

There are no changes inside the tests themselves, they just get wrapped in a timeout tracker.

While doing this, I also found a data race that gets triggered when the service is started with the cancellation token already cancelled. Fixed that also.


sebastianburckhardt:
With the added timeout checking we are seeing some definite failures in this CI so I will need to debug this.

sebastianburckhardt:
There seem to be bugs in the shutdown for the event processor. Since we are just about to replace that code (https://github.com/microsoft/durabletask-netherite/pull/385) I think it is not worth the effort to track this down. Instead, I will add the changes made in this PR to that PR and troubleshoot from there.

PR #423, opened 2024-09-27
Debug hangs in CI

sebastianburckhardt:
We have seen a lot of hangs recently in the CI, which seem to not be caused by recent changes but look a bit like errors in the test framework. I created this PR as a way to further investigate and hopefully fix these hangs.

In this first step, I am trying to understand hangs in the QueriesCopyToTail test which I have seen repeatedly in CI. The hangs seem to have something to do with the checkpoint injection mechanism used by the test. To narrow this down, we revise this state machine and add lots of tracing.

### DIFF
--- a/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
+++ b/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
@@ -2081,7 +2081,10 @@ namespace DurableTask.Netherite.Faster
                         break;
 
                     case WriteReason.CopyToTail:
-                        takeValueFromOutput = (output.Val != null); // we have observed that src.Val is null sometimes, so if present, we use output instead
+                        // we have empirically observed that src does sometimes not contain the correct value (is null)
+                        // we are not sure if this is a bug in FASTER or intended behavior
+                        // as a workaround for those situations, we are passing the source value in the output parameter, which seems to work o.k.
+                        takeValueFromOutput = (output.Val != null);
                         if (takeValueFromOutput)
                         {
                             this.cacheDebugger?.Record(key.Val, CacheDebugger.CacheEvent.SingleWriterCopyToTailFromOutput, output.Version, default, info.Address);
--- a/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
+++ b/src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
@@ -1828,6 +1828,7 @@ namespace DurableTask.Netherite.Faster
         public struct Output
         {
             public object Val;
+            public int Version;
 
             public TrackedObject Read(FasterKV store, string eventId)
             {
@@ -2024,6 +2025,7 @@ namespace DurableTask.Netherite.Faster
                     dst.Val = trackedObject;
                 }
 
+                dst.Version = src.Version;
                 this.stats.Read++;
                 return true;
             }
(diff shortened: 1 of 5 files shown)


sebastianburckhardt:
I am going to rerun the tests several times to see if I can flush out more CI failures.

test/DurableTask.Netherite.Tests/ConcurrentTestsFaster.cs
```diff
@@ -109,7 +109,7 @@ async Task RunAsync()
 
                     var tests = new List<(string, Task)>();
 
-                    foreach ((string name, Task task) in scenarios.StartAllScenarios(includeTimers: true, includeLarge: true))
+                    foreach ((string name, Task task) in scenarios.StartAllScenarios(includeTimers: false, includeLarge: true))
```
davidmrdavid, line 112: why did this change?

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
```diff
@@ -2076,7 +2081,15 @@ bool IFunctions<Key, Value, EffectTracker, Output, object>.SingleWriter(ref Key
...
-                        this.cacheDebugger?.Record(key.Val, CacheDebugger.CacheEvent.SingleWriterCopyToTail, src.Version, default, info.Address);
+                        takeValueFromOutput = (output.Val != null); // we have observed that src.Val is null sometimes, so if present, we use output instead
+                        if (takeValueFromOutput)
+                        {
+                            this.cacheDebugger?.Record(key.Val, CacheDebugger.CacheEvent.SingleWriterCopyToTailFromOutput, output.Version, default, info.Address);
+                        }
+                        else
+                        {
+                            this.cacheDebugger?.Record(key.Val, CacheDebugger.CacheEvent.SingleWriterCopyToTail, src.Version, default, info.Address);
+                        }
```
davidmrdavid, line 2092: I think this is a good place to elaborate a bit more on the significance of `src.Val` being null. Is this a bug in FASTER, or expected, or something else? Can you please add a small blurb about that here?

davidmrdavid (commented):
So if I understand correctly, there were issues in our model KV in FasterKV?

test/DurableTask.Netherite.Tests/ConcurrentTestsFaster.cs
sebastianburckhardt: I saw timer tests failing spuriously (because of large time delays leading to unexpected results) so I did not think it was worth keeping them in there.

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
sebastianburckhardt: That's sort of what the comment on line 2084 is supposed to be.

test/DurableTask.Netherite.Tests/ConcurrentTestsFaster.cs
davidmrdavid: If I understand correctly, you're seeing that timers make creating a testing spec difficult (as in it's difficult to assert what exactly will happen when). Ok, that makes sense

src/DurableTask.Netherite/StorageLayer/Faster/FasterKV.cs
davidmrdavid: I'm not sure that gives me enough context though. The way the comment is phrased ("we have observed...sometimes") makes me thing it is not expected that `src.Val` would be null, and if that's true then I think that's worth calling out explicitly. It would tell me whether or not there's something weird that FASTER is doing here

davidmrdavid (approved):

PR #425, opened 2024-10-02
increase major version (2.0.0) and update packages

sebastianburckhardt:
In preparation of the 2.0.0 release:

- increases the version number for all packages
- updates to the latest durable task dependencies
- updates most other packages to the latest versions
- removes the netcoreapp2.2 target which was a 'dummy' target that is no longer needed I think

### DIFF
--- a/src/DurableTask.Netherite/DurableTask.Netherite.csproj
+++ b/src/DurableTask.Netherite/DurableTask.Netherite.csproj
@@ -49,17 +49,17 @@
   </ItemGroup>
   
   <ItemGroup>
-    <PackageReference Include="Azure.Core" Version="1.38.0" />
-    <PackageReference Include="Azure.Data.Tables" Version="12.8.0" />
-    <PackageReference Include="Azure.Messaging.EventHubs" Version="5.11.2" />
-    <PackageReference Include="Azure.Messaging.EventHubs.Processor" Version="5.11.2" />
+    <PackageReference Include="Azure.Core" Version="1.43.0" />
+    <PackageReference Include="Azure.Data.Tables" Version="12.9.1" />
+    <PackageReference Include="Azure.Messaging.EventHubs" Version="5.11.5" />
+    <PackageReference Include="Azure.Messaging.EventHubs.Processor" Version="5.11.5" />
     <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="11.2.3" />
-    <PackageReference Include="Azure.Storage.Blobs" Version="12.16.0" />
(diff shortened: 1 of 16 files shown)


src/DurableTask.Netherite/DurableTask.Netherite.csproj
```diff
@@ -49,17 +49,17 @@
...
-    <PackageReference Include="Azure.Storage.Blobs" Version="12.16.0" />
-    <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.16.2" />
+    <PackageReference Include="Azure.Storage.Blobs" Version="12.22.1" />
+    <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.17.1" />
     <PackageReference Include="Microsoft.FASTER.Core" Version="2.6.5" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
-    <PackageReference Include="System.Threading.Channels" Version="4.7.1" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="8.0.0" PrivateAssets="All" />
+    <PackageReference Include="System.Threading.Channels" Version="6.0.0" />
```
davidmrdavid, line 62: Do we **need** to upgrade these 2? For maximum bundles compatibility (we'll be releasing this in an existing bundles feed) I'd rather not majorly increase dependencies that we don't strictly need. We'll get another chance, with more time, to see if it's safe to increase this once we consume DF WebJobs V3, so I'd prefer to delay these major version increases until then

davidmrdavid (commented):
Only 1 concern

src/DurableTask.Netherite/DurableTask.Netherite.csproj
sebastianburckhardt: ok, I will back those out

sebastianburckhardt: I am getting downgrade warnings if I don't include "System.Threading.Channels" Version="6.0.0" , must be being picked up somewhere

sebastianburckhardt: I think that:

- I can entirely remove reference to System.Threading.Channels because it is implicitly imported now by EH
- Sourcelink is a build-time only dependency so I don't think there is an issue with keeping it there.

davidmrdavid (approved):

Issue #430, opened 2024-10-10
ProcessEventsAsync Handler not set


Selekena:
We are setting up our .Net app with netherite against a storage account and eventhub. We get an exception when our app attempts to start. 

System.InvalidOperationException
  HResult=0x80131509
  Message=Cannot begin processing without the ProcessEventAsync handler set.
  Source=Azure.Messaging.EventHubs.Processor
  StackTrace:
   at Azure.Messaging.EventHubs.EventProcessorClient.<StartProcessingInternalAsync>d__73.MoveNext()
   at Azure.Messaging.EventHubs.EventProcessorClient.<StartProcessingAsync>d__54.MoveNext()
   at DurableTask.Netherite.EventHubsTransport.EventHubsPartitionManager.<StartHostingPartitionsAsync>d__16.MoveNext()
   at DurableTask.Netherite.EventHubsTransport.EventHubsTransport.<DurableTask-Netherite-ITransportLayer-StartWorkersAsync>d__39.MoveNext()
   at DurableTask.Netherite.NetheriteOrchestrationService.<StartWorkersAsync>d__94.MoveNext()
   at DurableTask.Netherite.NetheriteOrchestrationService.<>c__DisplayClass92_0`1.<<TryTransition>g__conditionalTransition|0>d.MoveNext()
   at DurableTask.Netherite.NetheriteOrchestrationService.<TryStartAsync>d__91.MoveNext()
   at DurableTask.Core.TaskHubWorker.<StartAsync>d__32.MoveNext()
   at Microsoft.Azure.EdgeProvisioning.Common.Workflow.Engine.WorkflowWorker.<StartAsync>d__10.MoveNext()
   at Microsoft.Azure.EdgeProvisioning.Common.Workflow.Engine.WorkflowEngine.<StartAsync>d__5.MoveNext()
   at Microsoft.Extensions.Hosting.Internal.Host.<StartAsync>d__12.MoveNext()
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.<RunAsync>d__4.MoveNext()
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.<RunAsync>d__4.MoveNext()
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.Run(IHost host)
   at Microsoft.AspNetCore.Builder.WebApplication.Run(String url)
   at Program.<Main>
  This exception was originally thrown at this call stack:
    [External Code]
[...]

PR #432, opened 2024-10-14
Fix missing ProcessPartitionEventAsync handler

sebastianburckhardt:
As observed in https://github.com/microsoft/durabletask-netherite/issues/430, the handler registration was missing.

This appears to be a simple copy-paste mistake but it was not caught by testing because the tests use the other constructor overload.


sebastianburckhardt:
As far as I know we don't really have a way to run automated tests using managed identity credentials.

Once we finish https://github.com/microsoft/durabletask-netherite/pull/433 we should at least be able to catch this using a smoke test, though.

PR #433, opened 2024-10-22
Support standard Azure Function configuration options for managed identity

sebastianburckhardt:
This PR adds support for configuring Managed Identity using the same configuration-section based approach that is used for other Azure Functions extensions. This makes it simpler to configure MI for .NET apps, and possible to configure it for all AF apps.   

For example, Azure Storage can be configured by defining `AzureWebJobsStorage__accountName`, `AzureWebJobsStorage__credential`, and `AzureWebJobsStorage__clientId`, and the Event Hubs namespace can be configured by defining `EventHubsConnection__fullyQualifiedNamespace`, `EventHubsConnection__credential`, and `EventHubsConnection__clientId`.

To avoid dealing with legacy credential types, this PR also replaces most remaining uses of the Microsoft.Azure.Storage.* (V11) SDKs with the Azure.Storage.* (V12) equivalents. Since this is not a major version bump, it does not remove the actual package dependency. Also, a few utility functions (name validation and connection string parsing) still depend on V11 since there is no equivalent in V12.

Remaining work:
- [x] run basic smoke tests for local and cloud deployment
- [ ] need documentation and perhaps a sample
- [ ] need to test runtime scaling


sebastianburckhardt:
Since we are doing a major version release for 3.0.0, I have spent a bit more time on this PR in order to remove all dependencies on the old Microsoft.Azure.Storage.Blob.

This required copying some utility code for validation and for connection string parsing which is no longer present in the new storage SDK.

cgillum (approved):

sebastianburckhardt:
remaining work items are tracked in https://github.com/microsoft/durabletask-netherite/issues/441

PR #437, opened 2024-11-20
Update Durable Dependencies Version for v3.0.0 Release

nytian:
As titled. This PR updates the below dependencies to start the Netherite v2.2.0 release:

Microsoft.Azure.WebJobs.Extensions.DurableTask 2.13.6 -> 3.0.0
Microsoft.Azure.DurableTask.Core 2.17.1 -> 3.0.0

### DIFF
--- a/src/DurableTask.Netherite.AzureFunctions/DurableTask.Netherite.AzureFunctions.csproj
+++ b/src/DurableTask.Netherite.AzureFunctions/DurableTask.Netherite.AzureFunctions.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>net6.0</TargetFrameworks>
     <LangVersion>latest</LangVersion>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <IncludeSymbols>true</IncludeSymbols>
@@ -45,8 +45,8 @@
   </ItemGroup>
   
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.17.1" />
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.13.6" />
+    <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="3.0.0" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="3.0.0" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/DurableTask.Netherite/DurableTask.Netherite.csproj
+++ b/src/DurableTask.Netherite/DurableTask.Netherite.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <LangVersion>latest</LangVersion>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <IncludeSymbols>true</IncludeSymbols>
--- a/src/DurableTask.Netherite/DurableTask.Netherite.csproj
+++ b/src/DurableTask.Netherite/DurableTask.Netherite.csproj
@@ -55,7 +55,7 @@
     <PackageReference Include="Azure.Messaging.EventHubs.Processor" Version="5.11.5" />
     <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="11.2.3" />
     <PackageReference Include="Azure.Storage.Blobs" Version="12.22.1" />
-    <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.17.1" />
+    <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="3.0.0" />
     <PackageReference Include="Microsoft.FASTER.Core" Version="2.6.5" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="8.0.0" PrivateAssets="All" />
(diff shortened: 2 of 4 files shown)


test/PerformanceTests/Benchmarks/Sequence/HttpTriggers.cs
```diff
@@ -4,16 +4,13 @@
...
+    using Microsoft.AspNetCore.Http;
     using Microsoft.AspNetCore.Mvc;
     using Microsoft.Azure.WebJobs;
+    using Microsoft.Azure.WebJobs.Extensions.DurableTask;
     using Microsoft.Azure.WebJobs.Extensions.Http;
-    using Microsoft.AspNetCore.Http;
     using Microsoft.Extensions.Logging;
-    using Newtonsoft.Json;
-    using Microsoft.Azure.WebJobs.Extensions.DurableTask;
-    using Microsoft.WindowsAzure.Storage.Queue;
```
nytian, line 16: These changes primarily about removing and sorting using, as Microsoft.WindowsAzure.Storage.Queue is no longer used with WebJobs.Extensions.DurableTask v3.

src/DurableTask.Netherite.AzureFunctions/DurableTask.Netherite.AzureFunctions.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>net6.0</TargetFrameworks>
```
nytian, line 4: I need to update the TFM here to successfully build, as DTFx.Core only supports .NET 6. However, I'm wondering if we should also consider updating the TFM for Netherite. @sebastianburckhardt . If not, we might need to consider some other ways here.

bachuv (approved):
Looks good to me! Please get @sebastianburckhardt's approval before merging though.

src/common.props
```diff
@@ -3,7 +3,7 @@
   <!-- Version settings: https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/ -->
   <PropertyGroup>
 	<MajorVersion>2</MajorVersion>
-	<MinorVersion>1</MinorVersion>
```
sebastianburckhardt, line 6: Don't we need to do a major release here? 

From my understanding (and I think @davidmrdavid mentioned this before) we cannot make a major version update of a dependency without also doing major version update of this package.

src/DurableTask.Netherite.AzureFunctions/DurableTask.Netherite.AzureFunctions.csproj
davidmrdavid: +1.
Put differently, it makes sense to have the TFM of the `Netherite.AzureFunctions` project match the "WebJobs extension".
And have the `DurableTask.Netherite` TFM match `DurableTask`

src/common.props
davidmrdavid: I agree. In-proc customers will have their DTF.Core (and DTFx.AS, due to the fact that the WebJobs extension imports the entire DTFx.AS backend _even when using Netherite_) version majorly upgraded.

Therefore, this should be `3.0.0`.

src/DurableTask.Netherite.AzureFunctions/DurableTask.Netherite.AzureFunctions.csproj
nytian: Right now [DurableTask.Netherite ](https://github.com/microsoft/durabletask-netherite/blob/main/src/DurableTask.Netherite/DurableTask.Netherite.csproj#L4)supports _netstandard2.1;netstandard2.0;netcoreapp3_. while [DurableTask.Core](https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/DurableTask.Core.csproj#L5) supports only netstandard2.0. I didn't update that yet in this PR, I am a little confused if I should update that too.

sebastianburckhardt: > I am a little confused if I should update that too.

I am not actually sure what the consequences of removing it are, i.e. who would be affected and how / why.

sebastianburckhardt: But maybe this is a good moment to remove unnecessary stuff since it a major version release.

nytian: I think we can leave it as is if you agree, given that targeting netstandard2.1;netstandard2.0;netcoreapp3.x is stricter and has a narrower scope compared to netstandard2.0. @sebastianburckhardt

sebastianburckhardt: I would be happy to try to replace `netstandard2.1;netstandard2.0;netcoreapp3.x` with just ` netstandard2.0` ... I don't think we need all of those. 

But I am uncomfortable doing that without some testing.

nytian: I can handle the testing. What specific tests do you think would be valuable, aside from the usual smoke and performance tests?

sebastianburckhardt: So my suggestion for now: change this to only `netstandard2.0`

And we will run a few tests (which we should do anyway) before the next release to make sure everything still works.

sebastianburckhardt: I have some additional load tests that are a bit more work (they are not fully automated) so I do only run them right before making a release. 

Are you planning to make a release soon?

nytian: Thanks. I am starting this mainly to catch up with the next functions bundles release. But if this takes time, I think it's ok to catch up the next bundles release. :)

nytian: Upgrade the TFM :)

src/DurableTask.Netherite/DurableTask.Netherite.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
nytian, line 4: update the TFM as suggested, this also aligns with the DTFx.Core v3.0.0

PR #444, opened 2025-02-05
Move to Central Package Management

nytian:
This PR moves to [Central Package Management](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management)

- Add Directory.package.props for CPM.

- Remove all versions at each csproj

### DIFF
--- a/samples/HelloDTFx/HelloDTFx/HelloDTFx.csproj
+++ b/samples/HelloDTFx/HelloDTFx/HelloDTFx.csproj
@@ -9,10 +9,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging.Console" />
-  </ItemGroup>
-
-  <ItemGroup>
-	  <ProjectReference Include="..\..\..\src\DurableTask.Netherite\DurableTask.Netherite.csproj" />
+    <PackageReference Include="Microsoft.Azure.DurableTask.Netherite"/>
   </ItemGroup>
 
 </Project>
--- a/samples/HelloDTFx/HelloDTFx/HelloDTFx.csproj
+++ b/samples/HelloDTFx/HelloDTFx/HelloDTFx.csproj
@@ -8,8 +8,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.DurableTask.Netherite" Version="1.4.2" />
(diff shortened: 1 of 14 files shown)


bachuv (approved):

sebastianburckhardt:
I am seeing a lot of warnings in the build output, like:
Found conflicts between different versions of "Microsoft.Bcl.AsyncInterfaces" that could not be resolved.
Is this related to the package management?

samples/HelloDTFx/HelloDTFx/HelloDTFx.csproj
```diff
@@ -8,8 +8,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.DurableTask.Netherite" Version="1.4.2" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" />
+  </ItemGroup>
+
+  <ItemGroup>
+	  <ProjectReference Include="..\..\..\src\DurableTask.Netherite\DurableTask.Netherite.csproj" />
```
sebastianburckhardt, line 15: The samples were designed to use the published package, not the locally built one. I suppose you are changing this so it works with centralized package management?

I am not sure what the reason was to use the published package. Probably so that the sample runs standalone and users can copy/paste it easily. That will of course no longer work with the centralized package management. But it's probably not a big issue.

sebastianburckhardt (approved):
Left some comments, otherwise LGTM.

samples/HelloDTFx/HelloDTFx/HelloDTFx.csproj
nytian: I can switch it back definitely. I thought we should make samples built on the latest Netherite and thus use the locally built one.

nytian: Updated. @sebastianburckhardt  As for the build version conflict warning of package "Microsoft.Bcl.AsyncInterfaces", is not becasue of the CPM, but because `Faster.Core` 2.6.5 is using v7.0.0, while other packages like `Azure.Core`,  sample packages `Microsoft.Extensions.Logging.Console` are still using v6.0.0.  I updated my PR to ping to use "Microsoft.Bcl.AsyncInterfaces" v7.0.0. to remove the warnings.

PR #451, opened 2025-06-16
Add explicit STJ reference for component governance

cgillum:
There are no direct references to System.Text.Json in this repo, but indirect references point to System.Text.Json 6.0.9, which has a known vulnerability. This change adds an explicit reference and sets the version to 6.0.10.


nytian (approved):

sebastianburckhardt (approved):
Overriding the version makes sense to me. Seems easier than chasing all the packages individually and checking for updates.

